PR #61, opened 2024-11-29
Start auto-generating neighbourhood boundaries

dabreegster:
This is an experimental start to generating boundaries automatically. It looks for 3 types of severances -- busy roads (based only on highway classification), railways, and bodies of water. Initial results are OK, but plenty of room for later improvement.

There are problems near the edge of the imported study area; we need to preserve more precise import boundaries to fix that.

We also need to prune the results to only include meaningfully neighbourhood-like things -- filtering out small traffic island areas or bits in between dual carriageways, bodies of water, etc. It'll probably depend on both geometric area / extent and semantics of what things border the area. Related to https://github.com/a-b-street/osm2streets/issues/248.


web/src/AutoBoundariesMode.svelte
```diff
@@ -0,0 +1,97 @@
...
+
+  let gj = JSON.parse($app!.renderAutoBoundaries());
+
+  function add(e: CustomEvent<LayerClickInfo>) {
+    let name = window.prompt("What do you want to name the neighbourhood?");
+    if (!name) {
+      return;
+    }
+    try {
+      // TODO Waypoints will be missing; editing will not work
```
dabreegster, line 23: Will be fixed by switching to the new area drawing tool, which is smart about handling freehand waypoints

PR #82, opened 2025-01-15
Pass click events "through" non-interactive 1-way layer

michaelkirk:
I was just kind of exploring the maplibre svelte library and randomly saw this `interactive = false`, which seems to solve a problem you mentioned the other day @dabreegster.

Your clicks will no longer be "intercepted" by the 1-way layer symbols.

**before:** you can see the hover effect turn off for the road every time you hover over a 1-way glyph.

**after:**

But oops! The cursor is no longer what we want, so I had to specify the cursor for the 1-way layer. It's weird, but it seems to work. Do you know if we use the 1way layer in any non-interactive contexts where we *should not* show the pointer cursor?

### DIFF
--- a/web/src/layers/OneWayLayer.svelte
+++ b/web/src/layers/OneWayLayer.svelte
@@ -1,6 +1,9 @@
 <script lang="ts">
   import { SymbolLayer } from "svelte-maplibre";
   import { layerId } from "../common";
+
+  // TODO Figure out if hoverCursor is necessary here, or if svelte-maplibre
+  // ignores it when interactive is false
 </script>
 
 <SymbolLayer
--- a/web/src/layers/OneWayLayer.svelte
+++ b/web/src/layers/OneWayLayer.svelte
@@ -21,4 +21,6 @@
   paint={{
     "icon-opacity": ["case", ["get", "direction_edited"], 1.0, 0.5],
   }}
+  interactive={false}
+  hoverCursor="pointer"
 />


dabreegster:
> Do you know if we use the 1way layer in any non-interactive contexts where we should not show the pointer cursor?

Looking at the uses:
- route mode: we don't want the cursor there, but trying it out, it's not happening there when I run locally...
- edit mode: yes (the one you looked at)
- debug mode: yes (you can click a road to open OSM)
- view shortcuts mode: yes when picking a road initially, no otherwise. Trying it there, it's also working fine as-is

Per https://github.com/dimfeld/svelte-maplibre/blob/dca839fe4951ac5a5ed3c3e666193f420589b92b/src/lib/Layer.svelte#L101, it looks like `hoverCursor` is ignored when `interactive=false`. I tried removing it, and things work as expected -- cursor stays as the pointer when hovering on roads and the one-way symbol. So I think just `interactive={false}` is the trick (great find!), and we don't need to worry about the cursor; other layers (that have the on:click behavior) set it, and it still takes effect.

michaelkirk:
Thanks for looking up the other uses for me. I should have done that myself.

> So I think just interactive={false} is the trick (great find!), and we don't need to worry about the cursor; other layers (that have the on:click behavior) set it, and it still takes effect.

I'm not sure I understand - are you saying I could revert the hoverCursor portion? Doesn't that contradict what my second video shows to be happening?

I can dig in further, I just want to make sure we're on the same page.

dabreegster:
> I'm not sure I understand - are you saying I could revert the hoverCursor portion? Doesn't that contradict what my second video shows to be happening?

... Huh. I tried again just now locally. For me, only with `interactive={false}` in OneWayLayer, the cursor still changes. To sanity check, I changed `hoverCursor="not-allowed"` in InteriorRoadLayer.svelte and carefully hovered over the arrow symbols. 

I can try with more browsers later and read through the svelte-maplibre code more closely. I can't reproduce the behavior in your second video.

dabreegster:
This PR moves us forward, so I'll merge it with a TODO to check into the behavior more carefully

Issue #36, opened 2024-03-02
Predict impact mode


dabreegster:
Some big changes requested here. I'd also like to rethink the demand model used for each area. And think through how to reasonably do this in WASM, or decide not to

dabreegster:
Screenshots of v1:

![Image](https://github.com/user-attachments/assets/f13cb4d4-0d4f-4f0e-ae51-bf155668d398)

![Image](https://github.com/user-attachments/assets/4b4fd5d2-49c5-45e8-b5e3-6189e5080bac)

michaelkirk:
Marking this as "taken" by you, but feel free to kick it back if you'er done working on it.

dabreegster:
![Image](https://github.com/user-attachments/assets/86a676f5-e60d-4637-b145-8a3c8902fd77)
From the Edinburgh CP plan
"Both widths & colours communicates most of the key info pretty well ie both the changes in traffic on each link AND the relative levels of each link across the network"

michaelkirk:
I think I like that visualization, but I have a clarification:

Obviously green means less traffic, red means more.

I **think** width means the relative change and the numbers floating along the road are the absolute difference in estimated trips. Can you confirm?

dabreegster:
From 3 April meeting: showing the absolute counts at all is risky / confusing, especially when comparing to model outputs from other tools. Can we completely remove these from display, or simplify into very broad buckets (0-1000, 1000-2000, etc)?

dabreegster:
For my own future reference, look for the email "Traffic Redistribution Visualisation" for more context here

Issue #77, opened 2025-01-12
Are edits tied to a neighbourhood boundary or not?


dabreegster:
Something quite weird today:

1) Draw a neighbourhood boundary
2) Change some filters, one-ways, etc inside there
3) Delete the boundary

The edits remain, but the user can't just go and click on an edited filter to delete it. Not sure how this should work.

Relatedly, from the main "pick neighbourhood" page, should there be a "clear all edits" option, or a summary of the edits?

dabreegster:
Notes from chat with @michaelkirk:
- the undo/redo feels broken now, because it crosses different neighbourhoods
- we could offer / force deleting all edits when deleting a boundary or changing it... except what if there's an overlapping other boundary that covers some/all?
- this probably isn't the most important thing to fix; a user can go edit something uneditable by drawing a new boundary around it

Issue #30, opened 2024-03-02
Edit turn restrictions


dabreegster:

dabreegster:
https://github.com/a-b-street/abstreet/discussions/1154
Once this feature is started, check into this bug that happened with the old version

PR #83, opened 2025-01-17
Start impact prediction mode

dabreegster:
A start to #36. After some changes to one neighbourhood in Edinburgh:
image
The impact is mostly localized around there:
image
Justifying why Queen Street got busier with examples:
image

### DIFF
--- a/backend/src/impact.rs
+++ b/backend/src/impact.rs
@@ -1,7 +1,7 @@
 use std::collections::HashMap;
 
 use geo::{Contains, Coord};
-use geojson::GeoJson;
+use geojson::FeatureCollection;
 
 use crate::{IntersectionID, MapModel, RoadID};
 
@@ -28,8 +28,9 @@ impl Impact {
         self.after_edits_counts.clear();
     }
 
-    /// Returns a feature per road, with `before` and `after` counts
-    pub fn recalculate(&mut self, map: &MapModel) -> GeoJson {
+    /// Returns a feature per road, with `before` and `after` counts, and a `max_count` foreign
+    /// member
+    pub fn recalculate(&mut self, map: &MapModel) -> FeatureCollection {
         if self.baseline_counts.is_empty() {
             info!("Calculating baseline impacts first");
             self.baseline_counts = map
@@ -49,17 +50,32 @@ impl Impact {
         }
 
         let mut features = Vec::new();
+        let mut max_count = 0;
         for road in &map.roads {
             let before = self.baseline_counts.get(&road.id).cloned().unwrap_or(0);
             let after = self.after_edits_counts.get(&road.id).cloned().unwrap_or(0);
-            if before > 0 || after > 0 {
+            max_count = max_count.max(before.max(after));
+            // Don't show unchanged roads
+            if before != after && (before > 0 || after > 0) {
                 let mut f = map.mercator.to_wgs84_gj(&road.linestring);
                 f.set_property("before", before);
                 f.set_property("after", after);
                 features.push(f);
             }
         }
-        GeoJson::from(features)
+
+        FeatureCollection {
+            features,
+            bbox: None,
+            foreign_members: Some(
+                serde_json::json!({
+                    "max_count": max_count,
+                })
+                .as_object()
+                .unwrap()
+                .clone(),
+            ),
+        }
     }
 }
 
@@ -67,7 +83,7 @@ impl Impact {
 fn synthetic_od_requests(map: &MapModel) -> Vec<(Coord, Coord)> {
     // TODO Or just directly use intersections and save the step of using closest_intersection?
 
-    let step_size_meters = 50;
+    let step_size_meters = 10;
     let boundary = map.mercator.to_mercator(&map.boundary_wgs84);
 
     let mut pts = Vec::new();
--- a/backend/src/impact.rs
+++ b/backend/src/impact.rs
@@ -58,8 +58,9 @@ impl Impact {
         for road in &map.roads {
             let before = self.counts_before.get(&road.id).cloned().unwrap_or(0);
             let after = self.counts_after.get(&road.id).cloned().unwrap_or(0);
+            // Don't show unchanged roads, but to scale the absolute counts, do look at the max
+            // count seen anywhere
             max_count = max_count.max(before.max(after));
-            // Don't show unchanged roads
             if before != after && (before > 0 || after > 0) {
                 let mut f = map.mercator.to_wgs84_gj(&road.linestring);
                 f.set_property("id", road.id.0);
--- a/backend/src/impact.rs
+++ b/backend/src/impact.rs
@@ -2,6 +2,7 @@ use std::collections::HashMap;
 
 use geo::{Contains, Coord};
 use geojson::{Feature, FeatureCollection};
+use nanorand::{Rng, WyRand};
 
 use crate::{IntersectionID, MapModel, RoadID};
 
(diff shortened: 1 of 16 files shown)


web/src/PredictImpactMode.svelte
dabreegster, line 1: I'm not really happy with any of the styling choices here yet, but it's a start and roughly recreates what's in v1. There's two things we probably want to communicate:

1) Did a particular road get more or less traffic than before edits? What roads changed the most? A diverging red/green color is used for this.
2) What's the level of traffic after, in absolute terms?

A road going from 10 trips to 50 trips seems like a dramatic change by color, but 50 trips total is nothing, so it should be thin.

backend/src/impact.rs
```diff
@@ -0,0 +1,126 @@
...
+        FeatureCollection {
+            features,
+            bbox: None,
+            foreign_members: Some(
+                serde_json::json!({
+                    "max_count": max_count,
+                })
+                .as_object()
+                .unwrap()
+                .clone(),
```
michaelkirk, line 82: minor nit: is this clone necessary - since it's the last thing in the method, I'd expect a move should be sufficient.

backend/src/impact.rs
```diff
@@ -0,0 +1,125 @@
...
+                f2.set_property("kind", "after");
+                changed_paths.push((f1, f2));
+            }
+        }
+
+        changed_paths
+    }
+}
+
+/// Deterministically produce a bunch of OD pairs, just as a fallback when there's no real data
```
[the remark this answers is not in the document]
michaelkirk: Seems like a reasonable approach for expediency. 

It might avoid trouble to regenerate trips where `i1 == i2`, but maybe it won't be an issue in practice.

backend/src/route.rs
```diff
@@ -115,6 +158,10 @@ impl Router {
             }
         }
         pts.dedup();
-        Some(LineString::new(pts))
+        LineString::new(pts)
+    }
+
+    pub fn crosses_road(&self, road: RoadID) -> bool {
```
michaelkirk, line 164: To make sure I'm understanding our terminology: a RoadID corresponds to a segment of the road network which has no intersections in its interior. Is that right?

So there's no way to "cross" a road without traveling along it. 

In common language, traveling "through" an intersection would mean traveling from one RoadID and then switching to another RoadID at the point of the intersection (even though in common language we'd call it the same "road" all along).

The above is all totally fine - just check if I'm misunderstanding.

michaelkirk (commented):
Embarrassing question, but how do I get to this view?

backend/src/lib.rs
```diff
@@ -294,6 +295,31 @@ impl LTN {
         .map_err(err_to_js)?)
     }
 
+    /// Returns GJ with a LineString per road, with before/after counts
```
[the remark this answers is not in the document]
dabreegster: > avoid ever doing application logic with geojson, and do everything with fully expressive rust types.

There's a bunch of other places in the code currently violating this, creating features and doing lots of `set_property`. The WASM layer in `lib.rs` is at least the one place to stringify JSON or parse inputs. It would indeed be nice to write structs using `geojson::ser` for most app-level code to output, especially if we could wire up something like https://docs.rs/ts-rs/latest/ts_rs/ to have more rigor about keeping `wasm.ts` in sync.

For the moment, I think the codebase is small/simple enough to keep playing it kind of loose, so I won't prioritize it yet.

backend/src/impact.rs
dabreegster: Sadly yes, [as_object](https://docs.rs/serde_json/latest/serde_json/enum.Value.html#method.as_object) returns a ref, and there's no `into_object` helper

dabreegster:
> Embarrassing question, but how do I get to this view?

image
From the pick neighbourhood page. I would add it to edit mode too, but the "other related actions" nav is already overflowing:
image

I think the small breadcrumb nav is useful in the top bar, but I'm not really sure where to put the "other actions". The left sidebars get busy too. Other people have had feedback that it's hard to see those actions on the top right.

And to be fair, the old version wasn't really that much better; the top was meant to be mutually exclusive modes, all of which can be jumped to whenever. And the bottom had a few specific to one neighbourhood:
image

I'll leave this alone for now, but ideas welcome

backend/src/route.rs
dabreegster: All correct. It will get a bit more complicated for #30, because the transition from one Road to another Road within an Intersection will have to be modeled explicitly. (Called "turn" or "movement" in abstreet.) Any ideas for more clear terms, happy to use

web/src/PredictImpactMode.svelte
```diff
@@ -0,0 +1,114 @@
...
+          <Link on:click={() => ($mode = { mode: "title", firstLoad: false })}>
+            Choose project
+          </Link>
+        </li>
+        <li>
+          <Link on:click={() => ($mode = { mode: "network" })}>
+            Pick neighbourhood
+          </Link>
+        </li>
+        <li>Predict impact mode</li>
```
michaelkirk, line 39:
```suggestion
        <li>Predict impact</li>
```

"mode" feels redundant in the UI

michaelkirk (approved):
> Embarrassing question, but how do I get to this view?

Thanks for the instructions! For transparency, I **did** look in the breadcrumbs, but I was drilled down into the neighborhood screen, not up a level at the "study area" screen.

I think it makes sense where it is, since we measure impact spanning *all* the neighborhoods.

I think mostly this speaks to me not having a good intuition yet for the data hierarchy.

Let me see if I have it right:

A "project" vs. "study area" are synonymous (or one-to-one).

vs. a neighborhood, which is a subset of a study area. A study area can have many neighborhoods.

When you are placing filters, you are drilled down to "editing a neighborhood", not editing the entire study area.

But because edits in one neighborhood might affect traffic in another, impact measurements must be done at the scale of the entire project/study-area.

Is all that correct?

Anyway, the changes look good, and seem like a good basis for this view. Thanks for walking me through it.

dabreegster:
Data hierarchy understanding is almost right! It is useful to see the confusing points here, because they're definitely confusing parts of the UX.

> A "project" vs. "study area" are synonymous (or one-to-one).

A study area is a place like Edinburgh or a section of Seattle. A project is meant to be like a file somebody makes (and later shares) to use the tool. The project is tied to one study area, but a study area can have multiple files.

Right now, the project has just one set of edits (changes to one-ways, modal filters, turn restrictions, speeds and a definition of neighbourhood boundaries). But in the old tool, there were multiple "proposals" per file/project, and a way to switch quickly between them. I'm not sure we need all that complexity here yet. Something closer to what I'm picturing is https://nptscot.github.io/npw/ (only a few places here have working files online -- Edinburgh or Glasgow, if you're going to try this). First the user restores a project file they were working on previously, or makes a new one by first picking the study area.

> When you are placing filters, you are drilled down to "editing a neighborhood", not editing the entire study area.

Correct, but this is probably confusing: #77

> But because edits in one neighborhood might affect traffic in another, impact measurements must be done at the scale of the entire project/study-area.

Exactly. The flow a user might take:

1) Start with a neighbourhood boundary they have in mind already, or use the future neighbourhood prioritization module to pick one.
2) Within the edit neighbourhood view, edit stuff to achieve low shortcuts and good cell splits. There's pretty instant feedback for this. The problem being solved is traffic cutting through this one area.
3) Afterwards, someone needs to understand if there could be unexpected spillover effects to nearby neighbourhoods. Plug the leak in one area, but just push traffic to a residential street in the adjacent neighbourhood? No good, need to think through that too. Or push traffic onto the main road (the expected short-term consequence, before "traffic dissipation" can occur) -- maybe there are bus routes there or other issues, and actually a filter on a main perimeter road could be worth thinking about.

--

Thanks for the review! I'll wait for feedback from upstream about "predict impact" vs "predict areawide traffic impacts" or something else before renaming.

web/src/PredictImpactMode.svelte
dabreegster: Done

Issue #86, opened 2023-08-22
LTN: option to make segment more car friendly


XioNoX:
A bit similar to a-b-street/abstreet#1029 but it's not actually a bug,

In my city there is a transit road that have a small "living street" section: https://www.openstreetmap.org/way/462948349

![Screenshot from 2023-08-22 21-21-16](https://github.com/a-b-street/abstreet/assets/688817/353a79ab-47d1-4ad7-bd2d-43d78c05a1ae)

Because of that, routing skips it fully and offload traffic to another street further away, which is not what local drivers do.

![Screenshot from 2023-08-22 21-31-20](https://github.com/a-b-street/abstreet/assets/688817/5e842b95-4f10-4097-88f4-427011402f02)
![Screenshot from 2023-08-22 21-37-27](https://github.com/a-b-street/abstreet/assets/688817/916c0973-dd03-4504-ad1b-4a82c246be5c)

Not sure what the best way to improve it is, but maybe add a button like "speed limits" to temporarily make it "car friendly" so simulations are more accurate.
More complex fix might be to ignore such usecase when a stretch of "living street" is alone on a larger road.

Thanks

dabreegster:
Ooh good catch, I meant to transfer all LTN issues from the old repo to this one

Issue #34, opened 2024-03-02
Route mode


dabreegster:
Parity with old version

- [x] Slow-down factor for main roads
- [x] More info in the panel

michaelkirk:
Closing as remaining work is captured in sub-issues.

Issue #28, opened 2024-03-02
Edit speed limits


dabreegster:
Is this so important? Revisit the original ask

michaelkirk:
As discussed in the call today, it seems we will:


1. We will not let the user edit speed limit at a per-segment level. That kind of detail is just too fine grained for the purposes of this tool.
2. We will not let the user explicitly edit speed limits at a per-neighbourhood level.
3. We will uniformly apply a max 20mph speed limit to every interior Road within the LTN's.
4. This new speed limit **will** be incorporated into our routing/impact model
5. Other than the effects visible in "impact mode" due to routing changes, we won't explicitly surface that the speed limit has been changed in the application.

Awkward: How to handle editing boundaries after the fact? I think it should be easy enough if we keep the "unedited roads" somewhere (do we?).

PR #87, opened 2025-01-18
Start handling speed limits properly

dabreegster:
Much of #28. Every road now has a speed, with backup guesses using the `highway` tag. There's a way to color roads by speed:
image
And routing now goes by time cost, not distance. Route mode and impact-to-one-place now plumb back distance and time info, that's displayed -- so bit of #34 too.

I'm not sure about the range of the color scale. It's lifted from a project focused on long-distance routes. LTNs are smaller; the difference between 20 and 10 mph feels more important.

Also no editing of these speeds yet. Not sure how the UI for that should look, and also a bit unsure what it semantically represents -- correcting basemap data, nuances of `living_street` (#86), or new traffic calming?


web/src/common/index.ts
```diff
@@ -34,3 +34,19 @@ export function roadLineWidth(extraWidth: number): ExpressionSpecification {
     24 + extraWidth,
   ];
 }
+
+export function prettyPrintDistance(meters: number): string {
+  if (meters < 1000.0) {
```
michaelkirk, line 39: I admit I'm completely ignorant, but I see all the speed limits are in miles. So should we be printing miles here?

michaelkirk:
Changes all seem reasonable to me. 👍

web/src/common/index.ts
dabreegster: https://www.reddit.com/r/AskABrit/comments/lcx4do/do_brits_use_miles/
... I went down a rabbit hole here. I guess I've just gotten a bit used to mixing systems here. Speeds are definitely mph, but I'm a bit unclear whether it's more conventional to use m/km or miles. Will check, thanks for flagging

Issue #41, opened 2024-03-02
Diagonal filters


dabreegster:
Parity with old version

dabreegster:
We should discuss before starting this one

TFCx:
Is it a very specific case of intersection customization ?

dabreegster:
In v1, you could click 4 way intersections to toggle:
![image](https://github.com/user-attachments/assets/5578bd0c-2168-481e-b664-d155208b8d3b)
![image](https://github.com/user-attachments/assets/5579f610-0792-499f-8f89-b6944d96f282)
These are used in the US sometimes. I think they can be helpful for grid layouts, with many 4-ways.

michaelkirk:
I was unable to apply a diagonal filter in v1. Any hint how/where to achieve this?

https://github.com/user-attachments/assets/6d5eb18e-239d-48ce-a261-f390c0f7b0d4

Also, from that video - I'm not sure what's happening when I click the intersection a second time - seemingly more trips are excluded, but there's no visual indication of what happens between the first and second click.

dabreegster:
You need to first change the filter type to the red "no entry":

![Image](https://github.com/user-attachments/assets/19d6edfa-5596-4910-8a78-638ea1432865)
Then you can click 4-ways:

![Image](https://github.com/user-attachments/assets/7a81f6e9-129d-404c-b91b-98298758a8ee)

In the old version, there was a request to default to the green symbol over the red. They mean exactly the same thing in the tool, but have different connotations when shown in a consultation. Unfortunately only the red "no entry" one reasonably shows the direction of the filter. Unsolved design problem. In the new version we're rotating the entire icon right now, but the green one and bus gates look pretty strange as a result. #66 is about figuring out something nicer to do -- maybe keep the walking/cycling icon always upright, and add some kind of rotated line underneath it (without implying that the line means "prohibited")? Or maybe the diagonal filter deserves its own special icon? Or maybe we don't use a circular maplibre symbol layer for diagonal filters at all, but instead a thin line (figuring out something to do at low zoom)?

michaelkirk:
Just to log my understanding after looking at the v1 code base and interface:

Filters placed on intersections are handled differently from filters placed mid-segment.

When it's a 4-way intersection, any filter is a Diagonal filter. You can click to toggle between the two possible orientations for the diagonal filter.

When it's a 3-way intersection, you can click to toggle which of the 3 roads is "no access" - the other two roads remain connected.

[...]

PR #93, opened 2025-01-19
Start modelling movements at intersections

dabreegster:
Part of #30 (with some TODOs there) and also necessary for #41. This adds a debug mode (from the "pick neighbourhood" page) to understand movements at each intersection. They're generated by looking at one-way directions and the simplest OSM turn restriction cases. Only used in this debug mode right now, not yet for routing.
image

It's going to be much easier to flesh this out with a better unit testing story. I'm thinking of manually drawing / clipping some test cases in JOSM, checking those in here. For movements, maybe the output can be expressed in the code as a list of movements between named roads, or maybe some GeoJSON / SVG output would be faster to understand.

### DIFF
--- a/backend/src/create.rs
+++ b/backend/src/create.rs
@@ -21,6 +21,8 @@ struct Osm {
     railways: Vec<LineString>,
     waterways: Vec<LineString>,
     barrier_nodes: BTreeSet<NodeID>,
+    // Only represent one case of restricted turns (from, to) on a particular node
+    turn_restrictions: HashMap<NodeID, Vec<(WayID, WayID)>>,
 }
 
 impl OsmReader for Osm {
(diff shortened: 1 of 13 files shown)


web/src/DebugIntersectionsMode.svelte
dabreegster, line 1: I accidentally pushed e99bfc06de87a212bb13d8217dfbde7bd7dc1c2d straight to main, which set this up

backend/src/create.rs
```diff
@@ -74,6 +76,55 @@ impl OsmReader for Osm {
...
+            for (role, member) in members {
+                match member {
+                    OsmID::Way(w) => {
+                        if role == "from" && from.is_none() {
+                            from = Some(*w);
+                        } else if role == "to" && to.is_none() {
+                            to = Some(*w);
+                        } else {
+                            // Some other case, bail out
+                            return;
```
michaelkirk, line 105: I'm not very familiar with this logic yet, but should these be `continue`?

backend/src/geo_helpers.rs
```diff
@@ -214,6 +214,20 @@ pub fn make_arrow(line: Line, thickness: f64, double_ended: bool) -> Option<Poly
     Some(Polygon::new(LineString::new(pts), Vec::new()))
 }
 
+pub fn thicken_line(line: Line, thickness: f64) -> Polygon {
+    let angle = angle_of_line(line);
+    Polygon::new(
+        LineString::new(vec![
+            euclidean_destination_coord(line.start, angle - 90.0, thickness * 0.5),
```
michaelkirk, line 221: I should put this in geo... unfortunately I have a big open [LineMeasure change in PR limbo](https://github.com/georust/geo/pull/1298) that I'd like to resolve first... I'll go kick that to see if it's still alive. 🤔

backend/src/movements.rs
```diff
@@ -0,0 +1,80 @@
+use geo::{Euclidean, Length, Line, LineInterpolatePoint, Point, Polygon};
+use geojson::GeoJson;
+
+use crate::{
+    geo_helpers::{make_arrow, thicken_line},
+    Direction, IntersectionID, MapModel, Road,
+};
+
+impl MapModel {
+    pub fn get_movements(&self, i: IntersectionID) -> GeoJson {
```
michaelkirk, line 10: No biggie so take it or leave it: `movement_arrows` or `movement_labels` might be a better name, since these seem to be intended only as UI affordances, vs some business-logic struct.

michaelkirk:
image

What am I looking at here... 🤔 I initially assumed that the red arrows are illegal moves and the blue one is legal, but that doesn't seem correct. I'd expect there would need to be at least 1 way to either enter or exit each segment.

Or maybe I am reading it correctly, and we just haven't fleshed out all the legal turns from the available restrictions yet?

michaelkirk (commented):
Code seems good, but I don't quite understand the overall change yet.

backend/src/movements.rs
dabreegster: To minimize churn, I'll leave this for now. Right now it's a debug mode, but that will probably turn into some kind of "edit turn restrictions at an intersection" mode, and the movements sent over will have a semantic ID that the backend knows how to edit.

PR #95, opened 2025-01-23
Fix layout of maplibre controls by only using pico in explicit parts of the layout.

michaelkirk:
picocss is conflicting with maplibre - most notably the maplibre
controls have a silly amount of padding, so the controls look wonky.

Instead of applying picocss everywhere, we now "opt in" specific subtrees of
the DOM with the "pico" class.

It's kind of annoying, but it seems like it has to be this way -
if we're going to have two strong design systems (pico and maplibre),
they need to be confined to their own respective domains.

Note: it should be possible to use picocss within a tooltip/popup or something within the map, but we'd need to add a "pico" class to it.

### screenshots

**before:** Note how the zoom in/zoom out and search field controls in the map are very long.

Screenshot 2025-01-22 at 16 21 48

**after:**

Screenshot 2025-01-22 at 16 21 45

### DIFF
--- a/web/src/App.svelte
+++ b/web/src/App.svelte
@@ -2,7 +2,7 @@
   import onewayArrowUrl from "../assets/arrow.png?url";
   import logo from "../assets/logo.svg?url";
   import About from "./About.svelte";
-  import "@picocss/pico/css/pico.jade.min.css";
+  import "@picocss/pico/css/pico.conditional.jade.min.css";
   import initLtn from "backend";
   import type { Map } from "maplibre-gl";
   import { init as initRouteSnapper } from "route-snapper-ts";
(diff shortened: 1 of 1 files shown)


michaelkirk:
note: initially the "help" button was rendering green (the default button style rather than the component specific style).

I think this is because with the introduction of the "pico" class qualification, there is now a "specificity tie" for some of our rules. The immediate work around would be to slap a class on anything ambiguous (e.g. class="help-btn"), but that feels redundant. 

Anyway, the problem has since gone away for me, and I'm hoping I was just seeing the result of a caching bug, but LMK if you reproduce it and I can try to find a more systematic solution.

dabreegster (approved):
Ahh, this is great, this has been annoying forever. I have some attempt in another repo to try and override the `maplibe-ctrl` button class or something and fix this, but that was the wrong approach. Conditional styles make sense and aren't verbose at all with the 3 slot system.

I'm not noticing any problems with the help button, so maybe that was temporary caching or something

PR #97, opened 2025-01-24
Make the relation between edits and neighbourhoods a bit more clear

dabreegster:
Addresses the obvious fixes from #77.

1) Internal code rename to match UI
2) When you initially load a savefile, the user now can't "undo" all of the edits in it
3) When switching neighbourhoods, the undo/redo lists clear, so the user can't accidentally modify something far away
4) Pick neighbourhood mode summarizes the edits
image


web/src/PickNeighbourhoodMode.svelte
```diff
@@ -79,6 +80,26 @@
     );
   }
 
+  function countEdits(gj: FeatureCollection): {
```
dabreegster, line 83: Being sloppier, we'd just have a mapping from string to a counter (auto-set to 0) per `kind` property, but I couldn't get this all nicely working with TS, so verbose for now

web/src/App.svelte
```diff
@@ -160,8 +160,8 @@
             paint={{ "fill-color": "black", "fill-opacity": 0.3 }}
           />
         </GeoJSON>
-        {#if $mode.mode == "network"}
-          <NetworkMode />
+        {#if $mode.mode == "pick-neighbourhood"}
+          <PickNeighbourhoodMode />
```
michaelkirk, line 164: Yay! better name!

web/src/App.svelte
```diff
@@ -90,7 +90,7 @@
 </script>
 
 <div class="pico">
-  <About class="pico" />
+  <About />
```
michaelkirk, line 93: whoops, thanks!

Issue #18, opened 2024-02-16
One-ways


dabreegster:
- [x] Detect existing oneways from OSM
- [x] Make shortcuts, routing, cells, etc respect this
- [x] Add controls to toggle
- [x] Figure out how to represent in edits. Same as modal filters, handle changes to basemap
- [x] Add to diff tests
- [x] Figure out how to visualize

dabreegster:
Not perfect, but more or less done

Issue #17, opened 2024-02-12
Feedback requested: overall UI design


dabreegster:
The design of the new tool is partly informed by feedback about the previous tool, but I'm mostly trying things out / flying blind without wireframes. If anybody would like to help wireframe things or offer design advice, I'd love the help.

![image](https://github.com/dabreegster/ltn/assets/1664407/52bc7e7b-b765-4db1-b87e-1cdc20939932)
This screenshot demonstrates a few patterns:
- The top bar has breadcrumbs for nav, showing the flow through the app. I think this is a big improvement over the old tool
- The top right has links to other modes. I don't think the styling is very clear right now, or if this is a good place for these links
- The help text per mode is very long. Maybe it should be a modal popup the first time you visit that page, or have a Help button somewhere, or be a collapsible card?

![image](https://github.com/dabreegster/ltn/assets/1664407/577bb338-9275-482e-b36e-2bfb8be27685)
The main edit mode is pretty clunky right now. You have to press a button before adding a filter, instead of just clicking a road. There will be more edit tools here eventually -- the one-way control and turn restrictions. Maybe something closer to the old tool would be better -- a bottom bar or control panel on the map, since the interactions are for the map.

# Translations / localization

I've never set this up for any project before, ever. From the technical side it doesn't look too bad, now that the frontend UI is just a Svelte app. https://github.com/kaisermann/svelte-i18n looks fine. For creating the translations, [Transifex](https://www.transifex.com/open-source/) and [Weblate](https://docs.weblate.org/en/latest/index.html) look popular -- have people used others?

A bigger question is about _what_ should be translated. Right now most of the text in the app is quite placeholder. The long paragraphs of help text are just a first cut. And depending where someone's trying to use this, maybe "LTN", "modal filter" etc aren't the relevant terms -- "liveable neighbourhood", "no-traffic zone", "superblock", etc might be more appropriate.

dabreegster:
Cleaning up issues, and I don't think this one is useful. People have been opening specific issues for feedback, and that's worked better

PR #98, opened 2025-01-30
Diagonal filters

### LINKED ISSUE #41: Diagonal filters

michaelkirk:
FIXES #41 

## Followup tasks - I'll create issues for these

-[ ] FIX: quickly rotating after adding DiagonalFilter causes zoom
-[ ] known issue: hover isn't re-shown after click until move
-[ ] routing based on turn restrictions and diagonal filters


backend/src/create.rs
```diff
@@ -166,6 +153,13 @@ pub fn create_from_osm(
         })
         .collect();
 
+    // Copy all the fields
+    let intersections: Vec<Intersection> = graph
+        .intersections
+        .into_values()
+        .map(|i| Intersection::from_graph(i, &roads))
```
michaelkirk, line 160: Intersection roads are now sorted clockwise from north, so this got changed around a little.

This was necessary to sanely "split" the roads coming into an intersection into groups of adjacent roads.

backend/src/geo_helpers/mod.rs
```diff
@@ -107,6 +107,37 @@ pub fn angle_of_line(line: Line) -> f64 {
     (line.dy()).atan2(line.dx()).to_degrees()
 }
 
+/// North is 0°
+/// East is 90°
+/// South  is 180°
+/// West is 270°
+pub fn euclidean_bearing(origin: Coord, destination: Coord) -> f64 {
```
michaelkirk, line 114: I was struggling a bit with the output range of `angle_of_line` so this simply transforms it to something I'm more familiar with. 

Is the output range of angle_of_line intentional?

web/src/ImpactDetailMode.svelte
```diff
@@ -5,6 +5,7 @@
   import { SplitComponent } from "svelte-utils/top_bar_layout";
   import { DotMarker, gjPosition, layerId, Link, PrevNext } from "./common";
   import { ModalFilterLayer } from "./layers";
+  import EditableIntersectionLayer from "./layers/EditableIntersectionLayer.svelte";
```
dabreegster, line 8: This is a read-only mode; the user shouldn't be able to click intersections and do anything here. Is this here to visualize the diagonal filters? If so, it looks like that's set up to grab data from `RenderNeighbourhoodOutput`. But this is a map-wide mode, so instead should we fold this into the `renderModalFilters()` WASM call or make one for diagonals?

web/src/common/Intersection.ts
```diff
@@ -0,0 +1,42 @@
...
+
+  get filterAngle(): number | undefined {
+    return this.feature.properties.filter?.angle;
+  }
+
+  get hasAlreadyRotatedFilter(): boolean {
+    if (this.filterAngle === undefined) {
+      return false;
+    }
+    // This is potentially incorrect, if the roads aren't (approximately) evenly distributed.
```
dabreegster, line 38: Ah, this might be the bug. If I force this to false always, then I can get to the second rotation state (but of course then not delete or keep toggling through the 3 states).

That `split_offset` in the backend was 0 or 1always, right? Could we track and plumb that along here to know which rotation state we're in maybe?

web/src/common/Intersection.ts
```diff
@@ -0,0 +1,42 @@
...
+};
+
+export type IntersectionProperties = {
+  kind: "editable_intersection";
+  intersection_id: number;
+  filter?: DiagonalFilter;
+};
+export type IntersectionFeature = Feature<Point, IntersectionProperties>;
+
+export class Intersection {
```
dabreegster, line 16: I've gotten into the habitof not making classes in all my Svelte projects, partly because Svelte 4 has some oddness with reactivity outside of `.svelte` files. Svelte 5 fixes all of that, but I haven't thought through how things might be written differently now.

This is fine here and nice to see; I'll think about how other parts of the code could be more organized.

web/src/common/index.ts
```diff
@@ -51,3 +54,30 @@ export function prettyPrintTime(seconds: number): string {
...
+ * This is useful (e.g.) for using symbol layers whose output reflects the size of a physical feature on the map.
+ *
+ * @param mapMeters The length on the map in meters
+ * @param minimumPixels A minimum size (in pixels) at all zoom levels. You can set it to 0 if you want the feature to disappear at low zoom levels.
+ */
+export function mapMetersToPixels(
+  mapMeters: number,
+  minimumPixels: number = 4,
+): DataDrivenPropertyValueSpecification<number> {
+  // Random googling gives me these values - I'm not entirely sure they're correct but they seem about right.
```
dabreegster, line 70: !!! I've been trying to figure out how to do something like this in maplibre since the beginning!! Do you remember what you searched for to get this to work? I thought these values depended on the latitude? But I just tried in both Bristol and Dallas, and the circle size (when zoomed in close enough to see 30 meters) looks fine in both.

I just tried this same function on line widths, and it appears to work! Seriously, this opens up a whole space of design possibilities. Lack of zoom-independent distances has meant I'm previously turning lines into thick polygons on the backend, just to use polygon layers in maplibre.

web/src/edit/EditIntersectionPopup.svelte
```diff
@@ -0,0 +1,41 @@
...
+        alt="Remove diagonal filter"
+        style="transform: rotate(45deg);"
+      />
+      Delete diagonal filter
+    {:else}
+      <div style="font-size: 200%; padding-bottom: 4px;" class="icon">↻</div>
+      <img
+        src={`${import.meta.env.BASE_URL}/filters/diagonal_filter_icon.png`}
+        alt="Rotate diagonal filter"
+        style="transform: rotate(45deg);"
```
dabreegster, line 22: Rotating the icon rocks, I love this. I was trying to find examples of 4-ways at odd angles and wondering if we could dynamically rotate the icon to match what'll happen on the map exactly, but that's just an idea for later.

web/src/layers/EditableIntersectionLayer.svelte
```diff
@@ -0,0 +1,97 @@
...
+  export let interactive: boolean = false;
+
+  /// NOTE: this takes the intersection's index in the neighborhood FeatureCollection, *not* the intersectionId!
+  function getIntersectionByFeatureIndex(
+    intersectionIdx: number,
+  ): Intersection {
+    // The Feature from maplibre has its nested properties serialized as a JSON string.
+    // see: https://github.com/dimfeld/svelte-maplibre/discussions/162#discussioncomment-9295264
+    //
+    // So we refetch the corresponding IntersectionFeature from the neighbourhood to get all it's properties.
```
dabreegster, line 30:
```suggestion
    // So we refetch the corresponding IntersectionFeature from the neighbourhood to get all its properties.
```

web/src/layers/EditableIntersectionLayer.svelte
```diff
@@ -0,0 +1,97 @@
...
+    return intersection;
+  }
+
+  function getIntersectionFromFeatures(features: any): Intersection {
+    // Note: this `id` is autoassigned by maplibre, and corresponds to the index of the feature in the neighbourhood.
+    return getIntersectionByFeatureIndex(features[0].id);
+  }
+</script>
+
+<!--
```
dabreegster, line 44: Just checked, this is a pre-existing problem clicking on a road too. If you click and really click again to delete the new filter, the map zooms in.

`map.doubleClickZoom.disable()` should fix this. But... we'd have to blanket disable this while in edit mode entirely. Not sure what the right approach is. It'd be nice to conditionally block this behavior, depending if a layer is double clicked, or the map is.

web/src/layers/EditableIntersectionLayer.svelte
```diff
@@ -0,0 +1,97 @@
...
+
+ Maybe because `event.cancelable == false`
+-->
+<CircleLayer
+  {...layerId("editable-intersections")}
+  filter={["==", ["get", "kind"], "editable_intersection"]}
+  paint={{
+    "circle-radius": mapMetersToPixels(30),
+    "circle-color": Style.mapFeature.hover.backgroundColor,
+    "circle-opacity": hoverStateFilter(0.0, 0.5),
```
dabreegster, line 60: I really like the style/feel of hovering on intersections, nice

web/src/layers/EditableIntersectionLayer.svelte
```diff
@@ -0,0 +1,97 @@
...
+    onClickIntersection(getIntersectionFromFeatures(event.detail.features))}
+>
+  <Popup openOn="hover" let:features>
+    <EditIntersectionPopup
+      intersection={getIntersectionFromFeatures(features)}
+    />
+  </Popup>
+</CircleLayer>
+
+<SymbolLayer
```
dabreegster, line 78: This is the part that's confusing me. This component is doing two jobs, letting people interactively click intersections and rendering diagonal filters. For route mode and impact modes, we only want to render diagonal filters -- and in fact, we want to show them over the entire map, not just one neighbourhood.

What do you think about making a new map-wide API call to render all diagonal filters, or folding it into `renderModalFilters`?

dabreegster (changes requested):
Super awesome!

I'm seeing a bug where clicking some intersections toggles between 2 states, not 3 -- adds a filter in one case, then clicking removes it, instead of switching to the other diagonal:

image
This is on that same Bristol map, a bit west at #18.21/51.456875/-2.557481. I've seen the same behavior at #17.88/55.956921/-3.192951 in Edinburgh. Will comment if I spot anything in the code that might be the problem

backend/src/map_model.rs
```diff
@@ -609,6 +724,68 @@ pub struct ModalFilter {
...
+}
+
+impl DiagonalFilter {
+    /// Precondition: Intersection must be a 4-way intersection
+    fn new(
+        intersection: &Intersection,
+        split_offset: usize,
+        map_model: &MapModel,
+    ) -> DiagonalFilter {
+        // TODO: need to validate topology - some 4-way intersections cannot support all orientations
```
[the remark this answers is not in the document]
michaelkirk: For example, if you were at the intersection of two one-way roads, I think there would be only one valid position for the filter.  Right?

There is some value in consistent UX though - maybe this is another case of just leaving it as is and "let the user do dumb things so long as they can see that it's dumb and undo it".

web/src/edit/EditIntersectionPopup.svelte
michaelkirk: We do have that exact angle - it's used when rendering on the map. I'll take a look at incorporating it here - TBH I don't think it matters much for this popup.

web/src/layers/EditableIntersectionLayer.svelte
michaelkirk: Yeah, I was hoping that there was something like `event.preventDefault()`, but as of yet I haven't found it.

web/src/layers/EditableIntersectionLayer.svelte
michaelkirk: I think I understand your concern.

To be clear, unless I'm mistaken, the layer is only editable when you're in `Editing` mode. Everywhere else it has `interactive={false}`, which is sufficient to make it readonly (unless there's a bug....).

But two things I need to fix:

1. `EditableIntersectionLayer` is a terrible name for a layer that's often not editable.
2. I want these to be visible whenever the modalFilters are visible - e.g. in the "pick-neighbourhood" mode, which it currently does not. I'll take a look at folding it into renderModalFilters.

web/src/common/Intersection.ts
michaelkirk: I feel bad for trying to merge this. I was pretty sure it was incorrect. Thanks for quickly finding a counter example.

web/src/layers/EditableIntersectionLayer.svelte
```diff
@@ -0,0 +1,97 @@
...
+  import EditIntersectionPopup from "../edit/EditIntersectionPopup.svelte";
+  import type { RenderNeighbourhoodOutput } from "../wasm";
+
+  export let neighbourhood: RenderNeighbourhoodOutput =
+    getContext("neighbourhoodGj");
+  export let onClickIntersection = (intersection: Intersection) => {};
+  export let interactive: boolean = false;
+
+  /// NOTE: this takes the intersection's index in the neighborhood FeatureCollection, *not* the intersectionId!
+  function getIntersectionByFeatureIndex(
```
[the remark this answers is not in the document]
michaelkirk: (I meant to leave this comment last night, but I think it corresponds to https://github.com/a-b-street/ltn/pull/98#discussion_r1935589285)

I agree - svelte (at least svelte4) doesn't really seem to like passing around "business logic classes" aka "models". I find it such a helpful concept that I'm a bit loathe to give it up, but it might be more work than it's worth.

web/src/common/index.ts
michaelkirk: This wasn't the original link, but it has some of the same numbers:

https://gis.stackexchange.com/a/340056

web/src/layers/EditableIntersectionLayer.svelte
michaelkirk: I think disabling double click entirely while in edit mode would be a cure worse than the disease. At least it's clear whats happening now. The user can zoom back out. 

It is annoying though. Are you OK to merge without addressing this?

backend/src/map_model.rs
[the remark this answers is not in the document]
dabreegster: I think for now, letting the user do the dumb thing is fine (and hopefully easier to implement!) With turn restrictions as well, or having a filter on one of the adjacent roads... there's actually quite a few cases of "a diagonal filter doesn't make sense here."

web/src/common/Intersection.ts
dabreegster: Thanks for writing a comment expressing doubt :)  Makes finding bugs really easy

web/src/layers/EditableIntersectionLayer.svelte
dabreegster: Fine to ignore this problem for now

web/src/layers/EditableIntersectionLayer.svelte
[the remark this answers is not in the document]
dabreegster: Yeah, the combination of Svelte 4 lacking reactivity in `.ts` files and MapLibre not plumbing around "real" GeoJSON is tough. I think the small "model" classes on the frontend would be great, but not sure how to nicely pull them off. Svelte 5 should solve one of the problems. Using a partial ID to then make a second WASM call to grab a "full" model object could be fine for the second.

backend/src/map_model.rs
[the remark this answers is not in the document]
michaelkirk: I've pushed up some better docs.

backend/src/map_model.rs
```diff
@@ -609,6 +724,68 @@ pub struct ModalFilter {
...
+    pub angle: f64,
+    pub group_a: Vec<RoadID>,
+    pub group_b: Vec<RoadID>,
+}
+
+impl DiagonalFilter {
+    /// Precondition: Intersection must be a 4-way intersection
+    fn new(
+        intersection: &Intersection,
+        split_offset: usize,
```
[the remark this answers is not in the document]
michaelkirk: This exercise has made me realize it should just be a boolean. If we ever want to support 6/8/100-way intersections in the future, we'll have to revisit more than this anyway.

web/src/common/Intersection.ts
michaelkirk: Fixed! Note this breaks the serialization format, so you'll have to delete/recreate any projects that have diagonal filters.

backend/src/map_model.rs
```diff
@@ -609,6 +724,68 @@ pub struct ModalFilter {
...
+    }
+
+    // `movement`: (from, to)
+    pub fn allows_movement(&self, movement: &(RoadID, RoadID)) -> bool {
+        let (from, to) = movement;
+
+        debug_assert!(self.group_a.contains(from) || self.group_b.contains(from));
+        debug_assert!(self.group_a.contains(to) || self.group_b.contains(to));
+
+        self.group_a.contains(from) && self.group_a.contains(to)
```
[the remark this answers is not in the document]
michaelkirk: Done.

web/src/layers/EditableIntersectionLayer.svelte
michaelkirk: Ok, I've pushed up some changes and I think this is as  you described now (which is much better!) 

"Some day" I'd like to introduce some kind of `struct SaveFile`. I thinking working with geojson::Features is making it a bit clumsy.

backend/src/map_model.rs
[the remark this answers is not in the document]
michaelkirk: I've done this.

michaelkirk (commented):
Ok, I think I've addressed all your comments. Please take another look @dabreegster.

I've left the commits unsquashed for review purposes, but I'd recommend squashing it all before merging.

web/src/edit/EditIntersectionPopup.svelte
dabreegster: Agreed, the popups are perfectly usable as they are -- thanks

web/src/layers/EditableIntersectionLayer.svelte
```diff
@@ -0,0 +1,71 @@
...
+    // So we refetch the corresponding IntersectionFeature from the neighbourhood to get all its properties.
+    let intersectionFeature = neighbourhood.features[intersectionIdx];
+    let intersection = new Intersection(
+      intersectionFeature as IntersectionFeature,
+    );
+    return intersection;
+  }
+
+  function getIntersectionFromFeatures(features: any): Intersection {
+    // Note: this `id` is autoassigned by maplibre, and corresponds to the index of the feature in the neighbourhood.
```
dabreegster, line 34: FWIW, we could control these in Rust by setting the GJ feature ID. But since this output mixes roads and intersections and the ID spaces overlap, we couldn't do anything much easier

dabreegster (approved):
Brilliant, thank you! I'll file an issue to track the few followups, just so they don't get lost

PR #100, opened 2025-02-07
update to latest (not yet released) geo

michaelkirk:
NOTE: the test fixture changes appear to be due to https://github.com/georust/geo/pull/1310,
which reversed the Polygon winding on the output from geo::boolean_operations.

Now the exterior is wound counter-clockwise and the interior clockwise.

This output convention matches what's used elsewhere in geo and also in
the geojson spec.

In practice many tools are tolerant of both, but it's probably better to
follow convention here.


michaelkirk (commented):
There shouldn't be any meaningful changes in behavior here. Just keeping abreast of API changes.

backend/Cargo.lock
```diff
@@ -1128,7 +1128,7 @@ checksum = "3354b9ac3fae1ff6755cb6db53683adb661634f67557942dea4facebec0fee4b"
 [[package]]
 name = "utils"
 version = "0.1.0"
-source = "git+https://github.com/a-b-street/utils#10d8bf23282fb601ffc3eeb78b5ea8b9b1acd496"
+source = "git+https://github.com/a-b-street/utils?branch=mkirk%2Fupdate-geo#0c4274485d6295c2e0f3c89d74282400e27449d3"
```
dabreegster, line 1131: I'd be fine merging https://github.com/a-b-street/utils/pull/3, unless there's a reason to wait. If you mark it ready for review, I could do that now, if you like

dabreegster (approved):

backend/Cargo.toml
```diff
@@ -28,7 +28,7 @@ rstar = "0.12.0"
 serde = "1.0.188"
 serde_json = "1.0.105"
 serde-wasm-bindgen = "0.6.0"
-utils = { git = "https://github.com/a-b-street/utils", features = ["serde"] }
+utils = { git = "https://github.com/a-b-street/utils", branch = "mkirk/update-geo", features = ["serde"] }
```
michaelkirk, line 31: Corresponds to https://github.com/a-b-street/utils/pull/3

Not sure how you want to sequence these merges, or if you care. LMK!

michaelkirk: Thanks for merging https://github.com/a-b-street/utils/pull/3 - I've removed this branch specification, which is no longer needed and updated Cargo.lock to the latest utils@main.

👍

backend/src/map_model.rs
```diff
@@ -229,16 +229,14 @@ impl MapModel {
         self.roads
             .iter()
             .min_by_key(|r| {
-                let diff1 = Line::new(
+                let diff1 = Euclidean.distance(
```
michaelkirk, line 232: This was just a matter of taste - measuring line.length is the same as the distance between two points.

PR #103, opened 2025-02-07
Set up study areas for Scotland and a new landing page

dabreegster:
This PR takes a few steps:

1) Generates OSM clips for study areas covering Scotland
2) Sets up a new landing page to pick one of those areas, mimicking https://nptscot.github.io/npw/
3) Adds support for multi-polygon study area boundaries, necessary because some local authority districts are islands

I'll ASAP merge this and ask for async feedback, for speed / because diffbasing gets tedious. The new landing page will be at https://a-b-street.github.io/ltn/cnt.html.

PR #105, opened 2025-02-07
Start baking in OD demand data for Scotland

dabreegster:
1) Adapt data sources and code from NPW to build an origin/destination dataset for each Scottish area.
2) For those areas, plumb it into the WASM app and use it for predict impact mode, instead of the random OD pattern.
3) Start a simple UI to debug that demand data.

[Screencast from 07-02-25 11:13:08.webm](https://github.com/user-attachments/assets/e9561ec3-f8aa-4f10-b32f-4d4949084c44)

This is another quick PR to get things going. Async feedback very very welcome. I've added more followup tasks in #36.


backend/src/od.rs
```diff
@@ -0,0 +1,148 @@
...
+/// Origin/destination demand data, representing driving trips between two zones.
+#[derive(Serialize, Deserialize)]
+pub struct DemandModel {
+    pub zones: Vec<Zone>,
+    // (zone1, zone2, count), with count being the number of trips from zone1 to zone2
+    pub desire_lines: Vec<(ZoneID, ZoneID, usize)>,
+}
+
+impl DemandModel {
+    /// Turn all of the zones into Mercator. Don't do this when originally building and serializing
```
dabreegster, line 21: The zone data is shipped in as WGS84, then turned into Mercator for use here. I thought about just doing this upfront and serializing as Mercator, but it's too risky, since we calculate the bounding box from OSM input and turn into mercator here. That probably differs from the bbox around the study area GJ boundaries.

backend/src/od.rs
```diff
@@ -0,0 +1,148 @@
...
+                        requests.push((i1, i2, trip_count));
+                    }
+                }
+            }
+        }
+        requests
+    }
+
+    pub fn to_gj(&self, map: &MapModel) -> GeoJson {
+        // Per (from, to) pair, how many trips?
```
dabreegster, line 77: I'm not necessarily set on this as the best representation for the debug UI, just a starter

data_prep/.gitignore
dabreegster, line 1: Argh, this is a bit messy, it could go in the tmp/ dir for consistency. I started this work first, doubled back and made #103, then rebased. Will clean up later.

data_prep/scotland/src/main.rs
```diff
@@ -0,0 +1,120 @@
...
+    boundary_wgs84: MultiPolygon,
+    zones: &BTreeMap<String, Zone>,
+) -> BTreeMap<String, ZoneID> {
+    let mut matches = BTreeSet::new();
+    let mercator = Mercator::from(boundary_wgs84.clone()).unwrap();
+    let boundary_mercator = mercator.to_mercator(&boundary_wgs84);
+
+    for zone in zones.values() {
+        let zone_mercator = mercator.to_mercator(&zone.geometry);
+        if boundary_mercator.intersects(&zone_mercator) {
```
dabreegster, line 110: This is brute-force, but still only takes 70s end-to-end for everywhere. As an offline process, not that bad. We could use rstar, but should be careful and probably use British National Grid.

web/src/DebugDemandMode.svelte
dabreegster, line 1: This is placeholder, just to initially prove the data is flowing through end-to-end. Something like the good ol' commuter patterns UI from A/B Street is a next step!

backend/src/lib.rs
```diff
@@ -51,6 +52,8 @@ impl LTN {
     #[wasm_bindgen(constructor)]
     pub fn new(
         input_bytes: &[u8],
+        // Option doesn't work; the caller should just pass in 0 bytes to mean empty
```
michaelkirk, line 55: A doc comment would be good to remind us over in javascript land.

backend/src/od.rs
```diff
@@ -0,0 +1,148 @@
...
+
+/// Deterministically produce a bunch of OD pairs, just as a fallback when there's no real data
+pub fn synthetic_od_requests(map: &MapModel) -> Vec<(IntersectionID, IntersectionID, usize)> {
+    let num_requests = 1_000;
+
+    let mut rng = WyRand::new_seed(42);
+    let mut requests = Vec::new();
+    for _ in 0..num_requests {
+        let i1 = IntersectionID(rng.generate_range(0..map.intersections.len()));
+        let i2 = IntersectionID(rng.generate_range(0..map.intersections.len()));
```
michaelkirk, line 144: Should we check that i1 != i2?

data_prep/scotland/Cargo.toml
```diff
@@ -0,0 +1,20 @@
...
+csv = "1.3.0"
+geo = "0.29.1"
+geojson = { git = "https://github.com/georust/geojson", features = ["geo-types"] }
+serde = "1.0.188"
+utils = { git = "https://github.com/a-b-street/utils" }
+
+[patch.crates-io]
+# use unreleased `geo` and `geo-types` to fix crash: https://github.com/a-b-street/ltn/issues/54
+# (awaiting release of https://github.com/georust/geo/pull/1279)
+geo = {  git = "https://github.com/georust/geo" }
```
michaelkirk, line 19: Not a big deal, but we might want to start a workspace so we can consolidate some of these dependency declarations. Or is there value in keeping them separate?

data_prep/scotland/src/main.rs
```diff
@@ -0,0 +1,120 @@
...
+        );
+        std::fs::write(path, bincode::serialize(&demand)?)?;
+    }
+
+    Ok(())
+}
+
+#[derive(Deserialize)]
+struct StudyArea {
+    #[serde(deserialize_with = "geojson::de::deserialize_geometry")]
```
michaelkirk, line 65: I was wondering if anyone ever used these. 🤔 ❤️

data_prep/scotland/src/main.rs
michaelkirk: Just to make sure I'm following, since this runs locally, we could use `geo::Transform` (`--feature=use-proj`), right?

web/src/title/loader.ts
```diff
@@ -45,51 +49,63 @@ export async function loadFromLocalStorage(key: string) {
...
 
-    return [bytes, boundary];
+    return [osmBytes, null, boundary];
   } else {
     console.log(`Grabbing from Overpass`);
     let study_area_boundary = gj.features.find(
       (f: Feature) => f.properties!.kind == "study_area_boundary",
     )!;
     let resp = await safeFetch(overpassQueryForPolygon(study_area_boundary));
     let bytes = await resp.arrayBuffer();
```
michaelkirk, line 107:
```suggestion
let osmBytes = ...
```

web/src/DebugDemandMode.svelte
michaelkirk: IIRC - that was my first ABStreet work! 🐤

data_prep/scotland/src/main.rs
dabreegster: Yep. Might be worth revisiting this, but for now, this run-once step isn't that slow, and the results of doing WGS84 polygon intersect queries seems correct

backend/src/lib.rs
dabreegster: `wasm.ts` wraps this to make a nicer TS API:

```
  constructor(
    osmInput: Uint8Array,
    demandInput: Uint8Array | undefined,
    boundary: Feature<Polygon>,
    studyAreaName: string | undefined,
  ) {
    this.inner = new LTN(
      osmInput,
      demandInput || new Uint8Array(),
      boundary,
      studyAreaName,
    );
  }
```

web/src/title/loader.ts
dabreegster: Looks like a typo/mistaken suggestion?

data_prep/scotland/Cargo.toml
dabreegster: #110 to track

backend/src/od.rs
dabreegster: Done in c6b5c67ad47b905ae3697c538baf0861f2e0c273

web/src/title/loader.ts
michaelkirk: Sorry - I was being too concise. It was a minor nit anyway, so I just pushed up a commit: bf22d137a7c219f0b2aabac85ab40f8a9b4448e9

Issue #106, opened 2025-02-07
CNT contextual layers


dabreegster:
#20 is about adding contextual layers to help the user both initially pick their neighbourhood boundary and do designing inside. That issue will be about implementing this generally for anywhere in the world, using OSM and other global datasets.

In the short term though, we're going to start this work focused just in Scotland. The initial list mostly comes from two existing projects that should be quick to start adapting here.

From https://nptscot.github.io/npw/ / https://nptscot.github.io/:

- [ ] GPs and hospitals
- [ ] Schools
- [ ] Greenspaces
- [ ] Existing cycle infrastructure (blindly trusting the NPT pmtiles file)
- [ ] NPT flows (with full filtering / styling controls)
- [ ] Estimated traffic volumes
- [ ] Gradient
- [ ] Current level of service (blindly trusting NPT data, not calculating it)
- [ ] SIMD aka deprived population
- [ ] Population density

From https://acteng.github.io/atip/browse.html:

- [ ] Railway stations
- [ ] Bus routes
- [ ] Bus stops, with frequency of use
- [ ] Stats19 collisions
- [ ] Car ownership

Issue #73, opened 2025-01-11
Explore other design systems


dabreegster:
Currently using https://picocss.com/, which was nice at first (gets out of the way and improves over lack of styles entirely.) But there might be nicer options out there. A few from some reading:

- https://www.beercss.com/, material, way more components, also pico style (just styles base HTML elements)
- https://picturepan2.github.io/spectre/index.html

Need to have another discussion with a client and their desired branding before changing anything.

michaelkirk:
It's also possible to do this stuff from scratch. Just putting it out there.

Even pico, which is purported to be minimal and to "get out of the way" has broken the layout in our maplibre buttons.

<img width="361" alt="Screenshot 2025-01-14 at 14 37 08" src="https://github.com/user-attachments/assets/727bf609-4fc1-4d89-8427-08f3ec2f4506" />

Though maybe we're just "holding it wrong".

dabreegster:
Yep, with modern CSS and how [Svelte scopes CSS](https://svelte.dev/tutorial/svelte/component-styles) to each component, the homemade approach could be fine. I've gone with pico under time constraints in the past, because it pushed other projects along beyond unstyled HTML, and have carried it forward here by default. Pico and maplibre indeed interact poorly -- the map controls, and also styles failing to apply inside popups without some effort.

Ideally soon, I'll figure out from upstream whether using/mimicking https://designsystem.gov.scot/ is important or not, and then we can do something more cohesive. In the meantime, feel more than free to improve how things are currently, including rethinking how controls are arranged entirely. For example, the v1 LTN tool made very different choices about the main edit controls:
![image](https://github.com/user-attachments/assets/06c3bb46-6fe0-42e1-b970-58eb470d1b08)
I started with text buttons here for implementation simplicity only.

dabreegster:
Duplicate of #404

PR #107, opened 2025-02-07
Copy over some layers from NPT project

dabreegster:
This starts #106, adding optional contextual layers that the user can toggle on anytime. These are lifted from https://github.com/nptscot/npw with minimal changes. The layers are hidden if the user isn't in a CNT area.

The controls are just about as painfully awkward as I could make them:
[Screencast from 07-02-25 16:38:28.webm](https://github.com/user-attachments/assets/5c6ad487-cd2c-4e24-b2b6-9d52103b7549)
MapLibre control styling is really fighting pico here! My intention here is to get lots of the work started quickly, and figure out how to polish the UI as we go along. There's some UX work happening in the parallel NPW project, and when it's ready in a few weeks, the same ideas could be applied here. In the meantime, please do suggest any ideas for improvements.


README.md
```diff
@@ -34,18 +34,18 @@ automatically use the latest OSM data. These are manually managed and hosted by
...
 choosing what study areas to cache:
 
 ```
 AREAS="bristol edinburgh strasbourg ut_dallas"
 
 cd web/public
-mkdir boundaries osm cnt_demand
+mkdir boundaries severance_pbfs
 
-cd osm
```
dabreegster, line 48: If you've got this cached locally, you'll need `mv osm severance_pbfs`. The naming is pretty awkward, but lots of other projects are currently pointing to it. I'll clean up my Cloudflare storage later and rename them all.

data_prep/scotland/get_reference_layers.sh
```diff
@@ -0,0 +1,57 @@
+#!/bin/bash
+
+set -e
+set -x
+OUT=out_layers
+mkdir -p $OUT
+
+function route_network {
+  # From https://github.com/nptscot/outputdata/releases
+  echo "Manually download https://github.com/nptscot/outputdata/releases/download/v2025-02-01/rnet_2025-02-01.pmtiles and save it to $OUT/route_network.pmtiles"
```
dabreegster, line 10: Private GH repo, sorry :(  But the data is public

data_prep/scotland/get_reference_layers.sh
dabreegster, line 1: These are all layers used in NPT/NPW. Totally TBD how to later generalize contextual layers outside of Scotland.

michaelkirk (approved):
> MapLibre control styling is really fighting pico here!

My impression so far is that we should not attempt to use pico within the context of maplibre. 

Granted it'd be nice to leverage some of their styles, but it's seemingly too opinionated (or maplibre is). Anyway, we can feasibly get rid of picocss, but I don't think we're getting rid of maplibre anytime soon.

I'm ok with merging it looking weird, but I'd like to follow up on this soon, and relatively quickly get a point to where we're making the app more presentable, not less, over time. 

I think saving all the "polish" for the end could hide some surprising complexity and also hinder user testing in the meanwhile. To be clear, I'm actually excited about doing that kind of work, so happy to have you punt it to me.

michaelkirk (changes requested):
Actually, I can't run this. Seems like maybe you forgot to `git add` a file?

After checking out this branch and doing: `(cd web && npm run wasm && npm run dev)`

Browsing to http://localhost:5173/ltn/ shows this in my js console:
```
SyntaxError: Importing binding name 'QualitativeLegend' is not found.
```

dabreegster:
> After checking out this branch and doing: (cd web && npm run wasm && npm run dev)

I updated svelte-utils in package-lock, so `npm ci` should do it

web/package-lock.json
```diff
@@ -2312,7 +2312,7 @@
     },
     "node_modules/svelte-utils": {
       "version": "0.0.1",
-      "resolved": "git+ssh://git@github.com/a-b-street/svelte-utils.git#ebe8244f143467088040aab01c86a19f13809f19",
+      "resolved": "git+ssh://git@github.com/a-b-street/svelte-utils.git#e9d4cefc8875bfe1ebfae7fa47de7574ec1c1c4f",
```
michaelkirk, line 2315: Ah, this should have been my cue to run npm ci.

dabreegster: I miss `cargo` and not worrying about this kind of thing :(

data_prep/scotland/get_reference_layers.sh
michaelkirk: Oh wait, I think I was confused. Based on your README updates, it looks like you've already done this by copying it onto od2net.org.

I've rebased #108 onto this branch to account for this and will push it up as soon as I've tested it.

README.md
dabreegster: (The new script needs to be updated too; if you merge first, I'm happy to rebase this and fix here

web/src/context/CBD.svelte
```diff
@@ -0,0 +1,141 @@
+<script lang="ts">
```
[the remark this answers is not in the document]
dabreegster: Clarified in c2586e3792a399042a2aec43a6a945843af69a4b -- cycling by design

dabreegster:
> I think saving all the "polish" for the end could hide some surprising complexity and also hinder user testing in the meanwhile. To be clear, I'm actually excited about doing that kind of work, so happy to have you punt it to me.

I agree completely. For me, it's much easier to get a chunk of functionality started, then figure out design later, because I generally struggle to make design choices. But absolutely leaving it for too long will make interim testing / feedback harder. 100% yes please to helping with this.

I still think pico is probably not going to survive long, but I'm not sure about the alternative (#73). In a few weeks, the parallel NPW project will have some professional design work done on it, and I bet many of the choices there can be adapted here. In the meantime, we should do anything to improve here.

https://actdev.cyipt.bike/dunton-hills/routes,buildings,collisions,trafficcounts,ltns/#12/51.576/0.38212 is maybe some inspiration for how to show multiple layers without being too much in the way, but I'm not sure it's appropriate here.
image

web/src/common/HelpButton.svelte
```diff
@@ -20,7 +20,7 @@
 
 <style>
   button {
-    background: none;
+    background: none !important;
```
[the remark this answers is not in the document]
dabreegster: Ah good catch, reverted

dabreegster:
PTAL. If you're OK with temporarily bad design, we can iterate later. If not, I'll try removing the `pico` class from ContextualLayers and getting things to look decent without it. Because we're under maplibre-ctrl classes too, there still might need to be more customizing to get it looking OK.

michaelkirk:
Yep, I'm ok to follow up with UI fixes in a separate PR (I've already created https://github.com/a-b-street/ltn/issues/109 to track it!).

I have to ask though, is this layout breakage intentional and part of the "needs cleanup", or is this surprising to you:

Before this branch, the map is alongside the choices:
Screenshot 2025-02-07 at 14 16 52

After this branch, the map is below the choices:
Screenshot 2025-02-07 at 14 30 30

I'm asking if you are aware of this, because I initially didn't see it. I think it appeared only after I cleared my node_modules cache (while investigating the help button `!important` css). I suspect there is some class of errors like this where there is a css class-specificity tie, so it's the ordering of the rules which is the tie-breaker determining which rule is picked, and (I'm guessing) the ordering of the rules is not stable with the hot-reload-dev-server thing.

Either way, if you want to punt it to me, I'm ready.

dabreegster:
> I suspect there is some class of errors like this where there is a css class-specificity tie, so it's the ordering of the rules which is the tie-breaker determining which rule is picked, and (I'm guessing) the ordering of the rules is not stable with the hot-reload-dev-server thing.

Aaaargh. No, this is not intentional, and quite frustrating. :(  I do not think I'll get to this before Sunday or Monday, so happy for any help.

michaelkirk:
As discussed out of band, we went ahead and merged this, and I'll follow up with any UI fallout.

Strangely, once deployed, the layout seems reasonable, but the green help button is back!

Screenshot 2025-02-07 at 14 58 50


My running theory is still something like this:

> I suspect there is some class of errors like this where there is a css class-specificity tie, so it's the ordering of the rules which is the tie-breaker determining which rule is picked, and (I'm guessing) the ordering of the rules is not stable with the hot-reload-dev-server thing.

Perhaps it applies to not just the dev-server - maybe the build process (vite? rollup?) are non deterministic in their ordering.

PR #108, opened 2025-02-07
Move the dev-data docs to an executable script

michaelkirk:

PR #113, opened 2025-02-07
Fix tests, CI fetches data with dev script

michaelkirk:
#107 changed where the pbfs live, but we didn't update the unit tests, so those are failing locally now.

Ironically, because we *also* didn't change the CI script to match our new developer layout, CI continued to pass.

In #108 I added a script to download all the useful dev data. I wanted to use that script in CI to fetch data in the same way we're using it locally, but the dev data is quite large (500MB+), so I've broken out a `fetch_test_data` subscript that fetches only the data we need for testing (20MB or so).

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -19,19 +19,7 @@ jobs:
           key: cargo-test-${{ hashFiles('backend/Cargo.lock') }}
 
       - name: Download test inputs
-        run: |
-          cd web/public
-
-          mkdir osm
-          cd osm
-          wget https://assets.od2net.org/severance_pbfs/bristol.pbf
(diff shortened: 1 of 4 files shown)


.github/workflows/test.yml
```diff
@@ -3,6 +3,7 @@ name: Tests
 on:
   push:
     branches: [main]
+  pull_request:
```
michaelkirk, line 6: Don't miss it! I also enabled tests for PRs. 👀👀👀

dabreegster (approved):
Thank you for cleaning up my mess! It's a code smell when a GH workflow starts to have more than a few commands; a separate script was a good idea for many reasons.

PR #114, opened 2025-02-09
Use a choropleth to show OD zone demand

dabreegster:
#36 
[Screencast from 09-02-25 14:26:17.webm](https://github.com/user-attachments/assets/1f0f150f-c2a6-4d97-aef8-f07f686e0122)
I'm mostly trying to mimick the good ol' commuter patterns from A/B Street. Does it seem reasonably understandable so far?

### DIFF
--- a/web/src/DebugDemandMode.svelte
+++ b/web/src/DebugDemandMode.svelte
@@ -48,8 +48,11 @@
         makeRamp(["get", key], limits, demandColorScale),
       ];
     } else {
-      // The arrays get stringified as properties and can't be used. Make an
-      // expression with a copy of the numbers.
+      // The GJ has nested arrays with the counts, but MapLibre stringifies
+      // these properties, so we can't use them for styling. Instead we
+      // dynamically make an expression to color each neighbourhood, by
+      // embedding the counts. The number of zones doesn't get too big,
+      // so this works well enough.
       let key: "counts_from" | "counts_to" = showTo
         ? "counts_to"
         : "counts_from";
@@ -103,6 +106,14 @@
       Trips to
     </label>
 
+    <p>
+      Trips {showTo ? "from" : "to"}
+      {hoveredId == null ? "each zone" : "this zone"}:
+    </p>
+    <SequentialLegend colorScale={demandColorScale} {limits} />
+
+    <hr />
+
     {#if current && hoveredId != null}
       <u>{current.properties.name}</u>
       <p>
@@ -118,13 +129,6 @@
     {:else}
       <p>Hover on a zone</p>
     {/if}
-
-    <hr />
-    <p>
-      Trips {showTo ? "from" : "to"}
-      {hoveredId == null ? "each zone" : "this zone"}:
-    </p>
-    <SequentialLegend colorScale={demandColorScale} {limits} />
   </div>
 
   <div slot="map">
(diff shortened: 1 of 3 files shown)


web/src/DebugDemandMode.svelte
```diff
@@ -1,31 +1,76 @@
...
+
+      let max = Math.max(...gj.features.map((f) => f.properties[key]));
+      let limits = maxIntoBuckets(max);
+
+      return [
+        limits.map((l) => l.toLocaleString()),
+        makeRamp(["get", key], limits, demandColorScale),
+      ];
+    } else {
+      // The arrays get stringified as properties and can't be used. Make an
```
dabreegster, line 51: The number of zones per study area isn't too huge, so this isn't horrible.

web/src/wasm.ts
```diff
@@ -207,14 +208,21 @@ export class Backend {
   }
 
   getDemandModel(): FeatureCollection<MultiPolygon, ZoneDemandProps> {
-    return JSON.parse(this.inner.getDemandModel());
+    let gj = JSON.parse(this.inner.getDemandModel());
+    for (let f of gj.features) {
```
dabreegster, line 212: An example where this WASM wrapper layer can do a bit of work, instead of having to do it on the Rust side

dabreegster (commented):
Ubuntu screen recording didn't capture my mouse moving, apologies if it's hard to interpret...

web/src/DebugDemandMode.svelte
michaelkirk: I'm assuming this has to do with maplibre stringifying any nested properties?

I'm not sure  I'm understanding this comment.

Is it: because we can't pass in an array of counts (due to maplibre nested properties limitations), we instead have to construct a verbose style rule that explicitly assigns the color to each neighborhood id.

web/src/DebugDemandMode.svelte
```diff
@@ -71,14 +118,21 @@
     {:else}
       <p>Hover on a zone</p>
     {/if}
+
+    <hr />
+    <p>
```
michaelkirk, line 123: Panning between neighborhoods causes the left column content to be really frenetic. 

I think moving the fixed height content (like the color ramp) above the other text would help,  or moving the color ramp over the map?

michaelkirk (approved):
> Ubuntu screen recording didn't capture my mouse moving, apologies if it's hard to interpret...

Indeed! It's pretty hard to follow what's going on without the mouse =(

I ran it locally and it seems reasonable. There's a couple things missing vs. the ABStreet commuter patterns:

1. click to "select" a neighborhood then you can hover over individual neighborhoods to see the value. I don't know that people actually care to see the *exact* value but I do think that selecting a "focused" neighborhood is helpful. For example, you can then toggle to/from.

2. The outgoing vs. incoming icon which is redundant with the To/From control, but maybe helpful to explain what exactly we're looking at here.

Since this is filed under "impact" - does there need to be a way to compare before vs. after treatments?

My feedback is all minor/additional. I think it's fine to merge this as is.

web/src/DebugDemandMode.svelte
dabreegster: Clarified in another commit. Your understanding is correct

web/src/DebugDemandMode.svelte
dabreegster: Good point... I moved the ramp above the other text. Maybe it should still be on the map, if we wind up consistently showing legends on the map.

Issue #101, opened 2025-02-07
Routing reflects turn restrictions (including diagonal filters)


michaelkirk:
Currently, routing only restricts movement based on whether roads are two-way or one-way.

We want to start accounting for turn restrictions. Turn restrictions might ultimately come from a variety of sources:

1. OSM input: e.g. `[type=restriction]` which correspond to existing on-the-ground conditions.
2. Diagonal filter's placed by the user (largely done in #41, translating this to turn restrictions should be straight forward 🤞)
3. Any manual per-intersection turn restriction editing by the user (TODO: #30 )

## Notes

Currently there are two places that build up a routing graph.

1. In [`Router`](https://github.com/a-b-street/ltn/blob/de1d2a95220d97c8cbac5b7624907f587cb941e0/backend/src/route.rs#L76) which is used for the Routes view.
2. In [`Shorcuts`](https://github.com/a-b-street/ltn/blob/de1d2a95220d97c8cbac5b7624907f587cb941e0/backend/src/shortcuts.rs#L51) which is used (probably among other things) to color the streets varying degrees of red to indicate "more cut-throughs". For me it's the most obvious indicator for where I should consider placing modal filters. 

Note that `Shortcuts` *also* uses `Router` (via `map.router_before` to measure impact), but `Shortcuts` indeed builds its own routing graph too.

The only difference in implementation I've seen between `Router`'s graph and `Shortcut`'s graph  is that the `Router` graph has a notion of "main_road_penalty" exposed to the user as "Slow down factor for main roads":

<img width="868" alt="Image" src="https://github.com/user-attachments/assets/5f875711-e22c-4009-adfd-1d9afca0670d" />

I'm not sure what that's about, but assuming it's useful. I don't think that will preclude sharing common route-graph-building code.

dabreegster:
> I'm not sure what that's about, but assuming it's useful. I don't think that will preclude sharing common route-graph-building code.

The two graphs have two differences:

1) That main road penalty. The edge costs for driving are just length / OSM speed limit currently. But the original motivation for LTNs is to stop shortcuts through residential areas. That problem happens more when the main/perimeter roads have more traffic. This is just a crude slider for "off-hours traffic" to "peak hour traffic". Increasing it should show more routes trying to cut through neighbourhoods.

[...]

PR #115, opened 2025-02-09
Start some minimal OSM unit tests

dabreegster:
For #6 and #101 and other reasons, it'd be very useful to test the LTN tool on really simple, minimal inputs. I'm proposing we:

1) Draw something simple in iD -- it's quite quick:
image
2) Write small bits of code to assert things, instead of a more cumbersome visual diff test (which just tries to check too many things at once)

If this approach looks good, I've got some cases related to detecting filters in #6 I'll turn into tests.

### DIFF
--- a/backend/src/osm_tests/README.md
+++ b/backend/src/osm_tests/README.md
@@ -13,4 +13,8 @@ Importing existing modal filters and turn restrictions from OSM can be very comp
 
 ## Viewing the OSM unit test inputs
 
-TODO
+To understand a unit test input:
+
+-  Just manually read it -- many are small enough to understand
+-  Load in <https://dabreegster.github.io/geojson-viewer/> -- but note relations don't show up
+-  Use JOSM
--- a/backend/src/osm_tests/README.md
+++ b/backend/src/osm_tests/README.md
@@ -1,0 +1,16 @@
+# OSM unit tests
+
+Importing existing modal filters and turn restrictions from OSM can be very complicated. This directory contains minimal synthetic OSM inputs to test different situations.
+
+## Creating a unit test
+
+1.  Open the iD editor and zoom in on an empty patch of ocean, [like here](https://www.openstreetmap.org/edit#map=17/55.704282/-0.110917)
+2.  Create the simplest example for some test case
+3.  Press **Save** in the top-right corner
+4.  Go to the bottom and **Download osmChange file**. Do **not** actually upload this edit!
+5.  Turn the .osc into osm.xml and add it to this directory: `osmium cat ~/Downloads/changes.osc -o my_test_case.osm.xml`
+6.  Add a test case in `mod.rs`, following examples there
+
+## Viewing the OSM unit test inputs
+
+TODO
--- a/backend/src/lib.rs
+++ b/backend/src/lib.rs
@@ -28,6 +28,8 @@ mod map_model;
 mod movements;
 mod neighbourhood;
 pub mod od;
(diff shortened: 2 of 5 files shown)


dabreegster (commented):
I also think we can start teasing apart the existing bigger tests. They've been quite noisy in recent PRs. We don't need to output existing modal filters in those tests; we can find every interesting situation, and express it as a smaller test here. The value of those tests could then be more about checking if roads are classified as perimeter/interior, checking the shortcut calculation, etc.

backend/src/osm_tests/README.md
```diff
@@ -0,0 +1,20 @@
+# OSM unit tests
+
+Importing existing modal filters and turn restrictions from OSM can be very complicated. This directory contains minimal synthetic OSM inputs to test different situations.
+
+## Creating a unit test
+
+1.  Open the iD editor and zoom in on an empty patch of ocean, [like here](https://www.openstreetmap.org/edit#map=17/55.704282/-0.110917)
```
michaelkirk, line 7: Thanks for writing this up! I was not clear on how we'd actually use this in practice. For some reason, it didn't occur to me to use an empty ocean.

backend/src/osm_tests/README.md
```diff
@@ -0,0 +1,20 @@
...
+4.  Go to the bottom and **Download osmChange file**. Do **not** actually upload this edit!
+5.  Turn the .osc into osm.xml and add it to this directory: `osmium cat ~/Downloads/changes.osc -o my_test_case.osm.xml`
+6.  Add a test case in `mod.rs`, following examples there
+
+## Viewing the OSM unit test inputs
+
+To understand a unit test input:
+
+-  Just manually read it -- many are small enough to understand
+-  Load in <https://dabreegster.github.io/geojson-viewer/> -- but note relations don't show up
```
michaelkirk, line 19: 😍 I didn't know you had an osm.xml viewer, nice! Can you share the source for it?

michaelkirk (approved):
This is great! I can definitely see myself using this.

> I also think we can start teasing apart the existing bigger tests. They've been quite noisy in recent PRs. We don't need to output existing modal filters in those tests; we can find every interesting situation, and express it as a smaller test here. The value of those tests could then be more about checking if roads are classified as perimeter/interior, checking the shortcut calculation, etc.

That makes sense. 

One other case I can imagine: For real world bug reports, I can imagine it being convenient to just grab a swathe of real world data from OSM that encompasses the problem (and probably some additional irrelevant context), and not being too worried about just checking the entire thing in, rather than always whittling it down to its simplest expression.

backend/src/osm_tests/README.md
dabreegster: It's a quick hacked-together GeoJSON viewing tool that winds up being faster for me to use than opening QGIS and figuring out how to symbolize by different properties. Has more property-level detail than geojson.io, and more free (and faster to load) than Felt. Adding OSM support was just gluing into a library that does the hard work:
https://github.com/dabreegster/geojson-viewer/commit/6fd5301a926bae56336a20c24a59ec0400b2ec30

michaelkirk: Dustin, you're writing javascript apps that aren't backed by rust. I'm growing concerned! 🙃 😉

dabreegster: I've done my penance with widgetry! Time to enjoy all the wonderful and only slightly confusing web technologies!

Issue #40, opened 2024-03-02
Fix basemap


dabreegster:
Per project.

Fix existing filters and one-ways. Relevant for the route mode's "before" and for predict impact. Potentially it'd also be useful to collect all of these and manually fix in OSM

dabreegster:
I think you're hitting some of the limits in #6. The first example uses a short `highway=cycleway` section, https://www.openstreetmap.org/way/983266180. This is correct OSM tagging AFAIK, but this case isn't handled yet here. Unfortunately I'm in the middle of a demanding paid project and can't work much on this until spring.

PR #119, opened 2025-02-11
routing reflects turn restrictions (including diagonal filters)

### LINKED ISSUE #101: Routing reflects turn restrictions (including diagonal filters)

michaelkirk:
Fixes #101 

Routing respects turn restrictions and diagonal filters. Internally, this required substantially changing our routing graph and many methods/data structures that implicitly relied on the previous structure.

Previously roads were the "edges" connecting "nodes" (intersections) in the routing graph. This was gratifying simple, and intuitively aligns with the geometry of roads as long things that connect points (intersections).

Unfortunately, in this previous model, any arrival "at" an intersection was treated equivalently - once you were "in" the intersection you could go anywhere else connected to the intersection.

In order to enforce turn restrictions, we need to know *how* you got to the intersection (from which road).

So previously...
Node: Intersection
Edge: Road

And now...
Node: (Road, Direction)
Edge: Intersection

== Tangential changes ==

- consolidate our router building logic so that the "whole map" routers   (e.g. `map_model.router_before`) and the neighbourhood router (`Shortcuts`) use the same construction logic.
- consolidate our "allowed movements" logic so that the debug-movements  screen is the same logic building our routing graph. I assume similar logic will power an eventual "edit turn restrictions" UI.
- Readability: replace is_forwards with named `Direction` enum, rename old
  `Direction` enum -> `TravelFlow`

== Perf ==

This significantly slows down router building and routing. I haven't investigated ways of improving this much, but for the record, here's the damage:

```
build router: bristol_east
                        time:   [21.873 ms 21.919 ms 21.975 ms]
                        change: [+233.23% +237.87% +240.77%] (p = 0.00 < 0.05)
                        Performance has regressed.

build router: strasbourg
                        time:   [442.47 ms 446.39 ms 450.27 ms]
                        change: [+431.14% +435.60% +439.89%] (p = 0.00 < 0.05)
                        Performance has regressed.

routing in bristol_east time:   [24.682 ms 26.817 ms 29.120 ms]
                        change: [+57.719% +74.930% +96.311%] (p = 0.00 < 0.05)
                        Performance has regressed.

routing in strasbourg   time:   [258.79 ms 260.86 ms 263.01 ms]
                        change: [+351.80% +374.31% +399.31%] (p = 0.00 < 0.05)
                        Performance has regressed.
```


backend/Cargo.toml
```diff
@@ -28,7 +28,8 @@ rstar = "0.12.0"
 serde = "1.0.188"
 serde_json = "1.0.105"
 serde-wasm-bindgen = "0.6.0"
-utils = { git = "https://github.com/a-b-street/utils", features = ["serde"] }
+#utils = { git = "https://github.com/a-b-street/utils", features = ["serde"] }
+utils = { path = "../../utils", features = ["serde"] }
```
michaelkirk, line 32: Note to self: PR to utils

backend/src/map_model.rs
```diff
@@ -834,24 +931,40 @@ impl FilterKind {
     }
 }
 
-#[derive(Clone, Copy, PartialEq)]
+#[derive(Clone, Debug, Copy, PartialEq, PartialOrd, Ord, Eq, Serialize, Deserialize)]
+pub enum TravelFlow {
+    BothWays,
+    OneWay(Direction),
+}
+impl TravelFlow {
```
michaelkirk, line 939: @dabreegster - this is kind of petty, so I wanted to check in. We were using a bool to represent `is_forwards` in a lot of places, even more with the new routing. 

To make it more self-documenting, I juggled things around, and renamed the "old" `Direction` -> `TravelFlow`.

Are you OK with this, or do you prefer the old way? I have a slight preference for the new way, but don't feel that strongly about it. If you're ok to move forward with this refactor, I'd like to proceed further and rename some fields/variables (from `direction(s)` -> `travel_flow(s)`) before marking as "ready for review".

backend/Cargo.toml
michaelkirk: Done: https://github.com/a-b-street/utils/pull/4

backend/src/map_model.rs
dabreegster: This new way is fine by me, feel free to finish renaming. The `is_forwards` bool was always kind of lazy/placeholder. In A/B Street, we had `DirectedRoadID(RoadID usize, bool forwards/backwards)` to make this cleaner.

backend/src/map_model.rs
```diff
@@ -89,6 +89,22 @@ pub struct Road {
     pub speed_mph: usize,
 }
 
+impl fmt::Debug for Road {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
```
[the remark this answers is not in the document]
dabreegster: This is much nicer, keep it!

backend/src/map_model.rs
```diff
@@ -124,8 +139,48 @@ impl Intersection {
...
+        router_input: &'a impl RouterInput,
+    ) -> impl Iterator<Item = (RoadID, Direction)> + 'a {
+        debug_assert!(
+            self.roads.contains(&from_r),
+            "{from_r:?} is not connected to intersection {self:?}"
+        );
+        let from_road = router_input.get_r(from_r);
+        self.roads.iter().filter_map(move |road_id| {
+            let to_road = router_input.get_r(*road_id);
+            if to_road.id == from_road.id {
```
dabreegster, line 154: I do not have any counter-examples and will have to think about how to force one, but this rule looks like it prohibits U-turns at a deadend. I don't think a route would ever choose to go down one, unless maybe there's a situation where it's forced to by a TR... ie:

```
       D
       |
  A--*--B
```

If the route starts from B, goes to *, there's a TR preventing B->A, then the route goes to D. It needs to u-turn there to make progress. Then southbound, D->A.

I don't think something this weird is likely to happen in practice, so don't block this PR on it. I've made a note to try and construct this test case later and see if it can actually occur; can file a followup issue if you agree it's possible

backend/src/map_model.rs
```diff
@@ -567,34 +622,94 @@ impl MapModel {
...
+                self.map.original_modal_filters.get(&r)
+            }
+
+            fn travel_flow(&self, r: RoadID) -> TravelFlow {
+                let road = self.get_r(r);
+                TravelFlow::from_osm(&road.tags)
+            }
+
+            fn diagonal_filter(&self, _i: IntersectionID) -> Option<&DiagonalFilter> {
+                // TODO: do we want to import pre-existing diagonal filters from OSM at some point?
```
dabreegster, line 652: I think they'll always be expressed in a different way, that might be kind of hard for us to "correct" to a simple 4-way intersection. Roanoke park is the only one in immediate memory:
https://www.openstreetmap.org/way/51538523
image
We'd have to detect this and stitch the two roads together somehow...

backend/src/map_model.rs
```diff
@@ -567,34 +622,94 @@ impl MapModel {
         Ok(())
     }
 
+    pub fn router_input_before(&self) -> impl RouterInput + use<'_> {
+        struct RouterInputBefore<'a> {
```
dabreegster, line 626: This is a cool pattern to get around the partial-borrow-of-some-struct-fields problem!

backend/src/map_model.rs
```diff
@@ -861,16 +984,16 @@ impl Direction {
     // TODO strum?
```
dabreegster, line 984: nit: you could remove this comment. I was considering a crate to autogen the to/from string, but it's overkill, especially now that this enum embeds a Direction

backend/src/route.rs
```diff
@@ -60,17 +84,23 @@ impl Router {
...
-                    input_graph.add_edge(i1, i2, cost);
-                    input_graph.add_edge(i2, i1, cost);
-                }
+
+            let mut link_through_intersection =
+                |intersection: &Intersection, direction: Direction| {
+                    let from = node_map.get_or_insert((road.id, direction));
+                    for outgoing_road in intersection.allowed_movements_from(road.id, router_input)
+                    {
+                        let to = node_map.get_or_insert(outgoing_road);
```
dabreegster, line 93: I almost got concerned here, but this is OK. In other projects using fast_paths, the initial contraction hierarchy (CH) gets rebuilt over time as the user edits something. Instead of building from scratch, it's sometimes faster to reuse the node ordering from the previous one. That requires the `NodeMap` to be exactly the same. For the representation in this PR, that would mean we need to carve out every (road, direction) pair in the NodeMap upfront.

But we're building from scratch every time (which is a perf issue we could improve on later), so for now, this is OK. Could be worth a brief comment here.

backend/src/route.rs
```diff
@@ -84,57 +114,48 @@ impl Router {
...
-        &self,
-        map: &MapModel,
-        start: IntersectionID,
-        end: IntersectionID,
-    ) -> Option<Route> {
-        let start = self.node_map.get(start)?;
-        let end = self.node_map.get(end)?;
-
+    pub fn route_from_roads(&self, start: RoadID, end: RoadID) -> Option<Route> {
+        info!("In Router::route_steps");
```
dabreegster, line 123: nit: remove or turn into debug

backend/src/route.rs
```diff
@@ -84,57 +100,54 @@ impl Router {
...
             return None;
         }
 
-        // TODO Reuse
+        // TODO Reuse: (Note `calc_path` is mutable)
         let mut path_calc = fast_paths::create_calculator(&self.ch);
-        let path = path_calc.calc_path(&self.ch, start, end)?;
+        let mut paths = vec![];
+        for start_direction in [Direction::Forwards, Direction::Backwards] {
+            for end_direction in [Direction::Forwards, Direction::Backwards] {
```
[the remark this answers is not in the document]
dabreegster: This is a little strange to think about, but I think you're right.

Impl-wise, this method can be faster and simpler by using https://github.com/easbar/fast_paths/?tab=readme-ov-file#calculating-paths-between-multiple-sources-and-targets. The extra weights for the different start/end combos can all be 0

backend/src/route.rs
```diff
@@ -177,3 +198,106 @@ impl Route {
...
+    use super::*;
+    use crate::osm_tests::load_osm_xml;
+
+    fn r(road_id: usize) -> RoadID {
+        RoadID(road_id)
+    }
+
+    #[test]
+    fn basic_route() {
+        // Helper to set up testing scenario. It's really verbose. We should probably do something else.
```
dabreegster, line 213: This comment can be removed, since the osm.xml tests are used now?

I like the ascii diagrams!

backend/src/shortcuts.rs
```diff
@@ -1,92 +1,65 @@
...
-            for end in &neighbourhood.border_intersections {
-                if start == end {
-                    continue;
-                }
-                if let (Some(i1), Some(i2)) = (node_map.get(*start), node_map.get(*end)) {
-                    if let Some(path) = path_calc.calc_path(&ch, i1, i2) {
-                        let mut steps = Vec::new();
+        for start_i in &neighbourhood.border_intersections {
+            let start_intersection = map.get_i(*start_i);
+            for start_r in &start_intersection.roads {
```
dabreegster, line 29: Something feels funny about the road-to-road routing and shortcuts... before, we found the shortest path between every pair of border intersections. Now we're doing it for every road connected to every border intersection. That doesn't seem intuitively right...

Using the shortcut viewer on an interior road adjacent to a border intersection, I think indeed this is wrong:
image
One of the shortcuts ends at a road outside the neighbourhood. I think we don't want to count that. The previous shortcut already captures going to the border intersection itself, because end_road = the highlighted interior road adjacent to the border:
image

I think the immediate fix is to make sure start and end roads are interior roads... or `editable_roads`, in case the user has selected perimeter roads. But I'm not sure about the second, for the same reasons in #124.

I'll keep thinking through the road-to-road routing implications, in case there's something simpler we're missing.

dabreegster (changes requested):
This is amazing! I think the only two blocking issues before merging this are:

1) using the fast_paths multi start/end to handle the 4 requests
2) filtering shortcuts to start/end on interior roads

And we can followup later on the rest

backend/src/map_model.rs
dabreegster: This was a pre-existing TODO that you just moved around, definitely fine to leave for later

backend/src/map_model.rs
dabreegster: (And to handle different input road lists, as you pointed out in another comment)

backend/src/map_model.rs
michaelkirk: I suppose this is something that could be mitigated somewhat by #40.

backend/src/route.rs
michaelkirk: Just to make sure I'm understanding - the comment should say something like:

> a given NodeId might refer to a different physical feature across rebuilds of the routing graph - do not assume they are stable

Is that what you mean?

Just to be clear, that behavior is not new to this PR.

[Previously](
https://github.com/a-b-street/ltn/pull/119/files/26665b3abfb3f3ecc60902ac63bdc0e79ade0028#diff-f2086e591ba396ff6f44c144fec9a8c43d1f5a44841c4cc098ee29b692fbd1c6L33), we skipped creating nodes when the road had a modal filter. If you added (or removed) a modal filter, that would similarly change the set of NodeIds.

backend/src/route.rs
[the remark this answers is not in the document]
michaelkirk: Oh nice! It gives a nice 20-30% reduction in call time for this method.

backend/src/map_model.rs
michaelkirk: I've created https://github.com/a-b-street/ltn/issues/137 to track it. My instinct is to wait and see if it comes up in practice to help determine how important this is to implement.

michaelkirk: To be clear though, I think you're correct: The proposed routing does not support u-turns.

backend/src/shortcuts.rs
michaelkirk: Your screenshots are identical, which confused me for a minute, but I get it now! Great catch!

> I think the immediate fix is to make sure start and end roads are interior roads

I've just now done that in https://github.com/a-b-street/ltn/pull/119/commits/f999c2bdd4d7f02fd1cba05d9b01aa12693dc83c

> ... or editable_roads, in case the user has selected perimeter roads.

We could do this... but does it actually *ever* make sense to be concerned about trips that are *strictly* on the perimeter? I would argue that that's not a shortcut at all. 

So, it's simpler and (I think?) more correct.  Also I think this might just resolve #124 - at least I can't think of an unhandled case.

michaelkirk:
Ok! I think I've addressed your concerns. Please take another look @dabreegster.

backend/src/shortcuts.rs
dabreegster: Argh, same screenshot got copied twice, sorry. Glad you figured it out, and this fix looks good.

dabreegster (approved):
This all works great, fantastic!

dabreegster:
Oh yeah, one thing to note -- because of direction -> travel_flow, old savefiles break. That's totally fine while we're still in this dev phase. I filed #138 to remind us to get stricter in a few months.

PR #139, opened 2025-02-13
Add missing loading screen when loading from url

### LINKED ISSUE #117: Landing page: Confusion after picking your study area
Follow up to #103 

The ux after picking your study area on the landing page is confusing. 

https://github.com/user-attachments/assets/37349e08-5e80-4699-a845-c8a547e156a3

- [x] It's weird to be hit with the splash screen **after** picking your study area. If anything, I guess it'd make more sense on `ltn/cnt.html`, before picking your study area, but  I actually think it'd be much better not to hijack the screen with a splash screen on load at all - it's very disruptive. I think users instinctually swat these away like a buzzing fly. How would you feel about putting this behind an "info/about" link?
- [ ] After you pick your study area, the map is zoomed out on the entire planet, rather than zoomed into my study area. I *think* what's happening here is some background loading. Obviously we should do what we can to make it faster, but if there's going to be any kind of delay, we should have some kind of activity indicator. In addition, if it's easy, while we're doing that work maybe we could zoom into the study area (do we have the bbox info at that point?) rather than zooming way out to the planet.

My guess is that these aren't surprising to you - probably it just represents an interstitial development state after copying in the landing page from another project and grafting it onto this one, but I wanted to document that I think something more needs to be done here.

michaelkirk:
FIXES #117 

This is also responsible for the loading screen after creating a new CNT project, since the CNT project redirects you from ltn/cnt.html -> ltn/index.html.

**before:**

https://github.com/user-attachments/assets/3b846510-2c67-4cfd-9ed8-c7b9b0f0cadd

**after:**

https://github.com/user-attachments/assets/1fb9d9dd-ff06-44b0-94f2-eb356efae1ef

I still think the relationship between ltn/cnt.html and ltn/index.html could be made clearer, but this is a quick fix that improves thing in any case.

### DIFF
--- a/web/src/title/TitleMode.svelte
+++ b/web/src/title/TitleMode.svelte
@@ -23,6 +23,7 @@
       let params = new URLSearchParams(window.location.search);
       let loadProject = params.get("project");
       if (loadProject) {
+        loading = `Loading project ${loadProject}`;
         loadFromLocalStorage(loadProject);
       }
     } else {


dabreegster:
Huzzah for quick wins!

PR #141, opened 2025-02-14
Add contextual layers for SIMD and population density.

dabreegster:
#106 
image
image

You'll need to grab  https://assets.od2net.org/cnt_layers/population.pmtiles. I mistakenly uploaded an early draft of it during the first layers PR, but it's changed now.

### DIFF
--- a/data_prep/scotland/get_reference_layers.sh
+++ b/data_prep/scotland/get_reference_layers.sh
@@ -65,6 +65,21 @@ function bus_routes {
   tippecanoe tmp/bus_routes_v2.geojson -zg -l bus_routes -o $OUT/bus_routes.pmtiles
 }
 
+function population {
+  # From https://www.data.gov.uk/dataset/1102bf85-ed49-440a-b211-da87e8d752eb/scottish-index-of-multiple-deprivation-simd-2020
+  download_to_subdir tmp https://maps.gov.scot/ATOM/shapefiles/SG_SIMD_2020.zip
+  cd tmp
+  unzip SG_SIMD_2020.zip
+  ogr2ogr population.geojson \
+          -t_srs EPSG:4326 \
+          SG_SIMD_2020.shp \
+          -nlt PROMOTE_TO_MULTI \
+          -sql 'SELECT DataZone as id, Rankv2 as imd_rank, Percentv2 as imd_percentile, SAPE2017 as population, OGR_GEOM_AREA as area FROM SG_SIMD_2020'
+  rm -f SG_SIMD_2020* SIMD2020v2*xlsx
+  cd ..
+  tippecanoe tmp/population.geojson -zg --generate-ids -l population -o $OUT/population.pmtiles
+}
+
 function get_scotland_osm {
   # Download Scotland OSM data
   if [ ! -f tmp/scotland-latest.osm.pbf ]; then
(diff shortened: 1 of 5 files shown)


web/src/context/Population.svelte
```diff
@@ -0,0 +1,129 @@
...
+  let showDensity = false;
+
+  // Color ramp from https://www.ons.gov.uk/census/maps/choropleth, dark to light.
+  let simdColorScale = ["#080C54", "#186290", "#1F9EB7", "#80C6A3", "#CDE594"];
+  // The percentiles are [1, 100]. The 5 colors cover 4 each.
+  let simdLimits = [0, 20, 40, 60, 80, 100];
+
+  let densityColorScale = simdColorScale.toReversed();
+  // Use the same (slightly rounded) buckets as https://www.ons.gov.uk/census/maps/choropleth/population/population-density/population-density/persons-per-square-kilometre. TODO Adapt for Scotland.
+  let densityLimits = [0, 4700, 13000, 33000, 94000, 1980000];
```
dabreegster, line 23: I lifted these from England, but since Scotland has lower density, we could maybe adjust

data_prep/scotland/get_reference_layers.sh
```diff
@@ -65,6 +65,21 @@ function bus_routes {
   tippecanoe tmp/bus_routes_v2.geojson -zg -l bus_routes -o $OUT/bus_routes.pmtiles
 }
 
+function population {
```
dabreegster, line 68: Lifted from https://github.com/nptscot/npw/blob/f7975400e052182b0b40e13685a01cb95e5a4b6e/data_prep/reference_layers.sh#L100

web/src/context/Population.svelte
```diff
@@ -0,0 +1,129 @@
...
+        target="_blank"
+      >
+        2020 data
+      </a>
+      .
+    </p>
+  </HelpButton>
+{/if}
+
+<VectorTileSource
```
michaelkirk, line 65: This is not novel to *this* PR, but I just realized I didn't have a good mental model for how our maplibre-svelte integration works.

The current component context for this is within the `Control` component defined in `ContextualLayers.svelte`, which is where the buttons and text like "Darker colours are denser" are rendered. 

However, the **layer** declarations below are just "floating" here, and somehow communicate to the outer map component. 

This is all fine and reasonable, I'm just restating my new understanding of how this works since I guess I hadn't paid enough attention before.

michaelkirk (approved):
LGTM

web/src/context/Population.svelte
dabreegster: It takes getting used to for sure.
https://github.com/a-b-street/ltn/blob/8f49d4f8bd9c54a0b6b833449040bae0995bf09f/web/src/App.svelte#L163
Because ContextualLayers is a child component of the MapLibre component, it all works out. svelte-maplibe sets context (the Svelte scoped state that's available to children, no matter how deep, without prop drilling) that lets sources/layers find the MapLibre map object.

It is kind of weird to mix DOM nodes and maplibre drawing directives. It's a nice way to write the code, because you can have one component control both the map and panel/legend, without plumbing state back and forth or doing something intense with event handlers. In this case, it works out easily, because there's a svelte-maplibre Control component for the floating map panels with bits of regular DOM.

In other cases, to get one svelte component to do something both on the sidebar and map, we use the SplitComponent trick. I can describe how it works if the code isn't clear. This is maybe also called "remote DOM" in other frameworks.

Issue #126, opened 2025-02-12
Visualize turn restrictions


dabreegster:
In v1, this was done with icons with "No X turn" symbols

dabreegster:
Started in `tr_icons` branch.

Design question 1: can we get away with one icon for this, using the angle to be clear? ie for no left turn:=
![Image](https://github.com/user-attachments/assets/33bcde71-d076-40e4-93a6-6423860dfecc)
An alternative is to have a different icon for left/right/straight/U-turn, and ignore the exact angle between roads. From v1:
![Image](https://github.com/user-attachments/assets/1076ad5b-504d-42c7-99ce-7a7da65881f3)


Design question 2: how do we show multiple restrictions originating from the same road? I'm thinking of vertically stacking the icons down the road, starting from the intersection where they apply and moving down. I could imagine something fancier like:
- showing multiple arrows in the one icon, always at the same position
- when you hover on this icon, expand icons perpendicular to the road (probably covering up other stuff) to show all cases

Design question 3: Are the "no U-turn" restrictions too noisy? Both of these examples are generally implied at dual carriageway joins/splits like this in the UK.
![Image](https://github.com/user-attachments/assets/2e53516c-290f-47a2-bc77-c6573a67fdac)

dabreegster:
An example of why I think we want the left/right/straight icons specifically: 

![Image](https://github.com/user-attachments/assets/647a8ef1-724b-45d5-86d8-e75fa5c671e7)
This is a no right turn to Elephant Rd, but from a bit farther away, it almost looks like the main road can't go northbound:

![Image](https://github.com/user-attachments/assets/3b0e4658-600f-4ffa-b37a-c8b2bfc5b337)

Another nice reference site I found, https://ahorn.lima-city.de/tr/, uses more specific icons and I think this is indeed the way:

![Image](https://github.com/user-attachments/assets/81776346-7a98-45e0-9b51-a5cb9d26f9ea)

TFCx:
Design 1 is more general and allow for 5/6/7/... branches intersections. And you could stack the one forbidden icon + 5/6/7 arrows to represent all the directions that are forbidden.

Could you make the turn restriction icon parametrized depending on the zoom ? Anyway, I think it needs a "two-pieces" (piecewise) black segment to better indicate from where the flow comes.

By putting the icon centered on the junction, it's relatively intuitive but i have concerns of readability in the case of multiple restrictions (i.e. originating from several upstream roads).

[...]

Issue #19, opened 2024-03-01
Permit changes on streets bordering areas


XioNoX:
Hi,

If for example we take that street : https://www.openstreetmap.org/relation/2233537

![Screenshot 2024-03-01 at 12-49-47 LTN from scratch](https://github.com/dabreegster/ltn/assets/688817/50aa6de3-b412-4faf-b747-f5e6ace69ebc)

It's currently not possible to set modal or one way filters on it as it's a zone edge. In most cases that would be fine as study areas are usually bordered by transit roads, but here the road is along a cliff. I also can't "cheat" the tool by selecting a larger study area.

Would it be possible to change that behavior ?

Thanks

dabreegster:
This feature request makes perfect sense to me, yes. I need to maybe revisit the implementation of discovering the perimeter roads to make it work, but I'll try and look at this soon. Thanks for the idea!

dabreegster:
![image](https://github.com/user-attachments/assets/5cfea25f-1a26-4c6c-a7de-dd8b4aa9ca86)
There's a new option to include perimeter roads for editing / shortcuts. This area is indeed tricky because of the cliff. When drawing the boundary, you can now also partly draw some freehand points if you need to stretch over more roads.

PR #146, opened 2025-02-16
Begin visualizing turn restrictions

dabreegster:
This is a start to #126. There are some problems with visualizing the TR icons, but I think they could be incrementally fixed up. If so, I'll move the tasks below to 126.

- [ ] When detecting TRs from OSM upfront, filter them out if they're redundant with a one-way. If a user wants to change this, they shouldn't have to take two actions, and the one-way is "stronger" than a TR, since it covers all source roads.
- [ ] Construct a unit test for the above. Also add another unit test based on https://www.openstreetmap.org/relation/15006809, where there's both a oneway and TR, but the TR is not redundant.
- [x] When hovering on a TR icon, this PR draws the movement arrow to be really clear about it. I'm still not sure this design is best. Maybe we should highlight both the "from" and "to" roads, maybe with two arrows back to back?
- [x] Some roads have multiple TRs originating from them. Right now the icons just brokenly overlap. `#19/51.4802457/-0.1080482` is a good test case. Stack them down the road? How to deal with really short roads or a big cluster of nearby TRs?
- [x] The rotation of the TR icons and the classification as left/right/straight/U turns is totally wrong right now. For classifying, I think we need to look at the two angles of the roads involved and a rotation between them?
- [ ] Consider de-emphasizing or hiding U-turns on dual carriageways. They're implied and probably easily understood by most users, so they just become noise.

On that last classification point, I considered using the OSM [restriction type](https://wiki.openstreetmap.org/wiki/Key:restriction). But:
1) it might be missing
2) it might be incorrect
3) when we add new TRs, we wouldn't have it, but we need it
Thus, I think we need to calculate it ourselves in any case.


michaelkirk (commented):
Seems like good progress!

It's a complicated set of functionality, with a lot of remaining work, so it's a little hard to understand some of what you're envisioning here. But I think:

1. Showing turn restrictions seems to have the signs upside down? I haven't done a thorough audit, but it seems like some (all?) of these are 180 degrees off. Maybe that's a simple typo, or related to my discussion further down about relative bearing vs. absolute heading.

2. To create a turn restriction, first you click on the "from", and then the "to" road. That seems like a reasonable first stab at a UI. Just spit-balling some UI affordances, maybe after selecting the "from" road,  when hovering over a candidate "to" road, we could have the cursor icon reflect what kind of turn restriction (left/right/straight) that *would* be placed.

3. re: moving the controls onto the map

I can see how it was getting crowded over there in the panel. And I'll be honest, my first impression of the buttons over the map is that it looks pretty good. But I have kind of a big monitor. On a smaller screen, I think this approach has less flexibility than before, and I think a crowded map would be worse than a crowded panel.

A potential upside of this change is that the tools are closer to the map content, so users don't have to move their mouse as far to switch between tools - but I'm not sure how relevant of an issue that is for people. It's certainly an important thing in an app like photoshop where people are doing lots of little touchups and switching tools back and forth a lot, but less so in a tool like this.

A downside of the move - some of the tools have context controls (e.g. what *type* of modal filter you want to pick seems contextually relevant to the "add modal" button, IMO that was better before when it was in the same panel.) I'm not yet sure if they'll be any context specific options for the turn restrictions stuff or other road editing like speed limit.

So, I have some concerns! But can you say more about why you like this change from a design standpoint? 

If it's mostly about uncrowding the tool panel, I think we could find another solution.

One dead simple thing would be to just render the controls in a vertical list, rather than a horizontal. Though I don't love that solution.

A slightly more complex one that I like more, we could organize our tools so that "draw many" becomes a context dependent sub-option of "Add Modal" rather than a top level item. 

Conceptually something like this:

## [x] Modal Filters
- `[ Dropdown: What kind]`
- [x] Add One (radio button-ish behavior)
- [ ] Add many along a line (radio button-ish behavior)

## [ ] Road Direction
- Click to reverse
- Maybe a good spot for speed limit controls

## [ ] Turn Restrictions
- Click "From"
- Click "To"
- ??? anything else here?

web/public/filters/README.md
```diff
@@ -1,3 +1,5 @@
 The photos were taken around London by Dustin Carlino in October 2022, and are
 released under
 [CC0](https://creativecommons.org/share-your-work/public-domain/cc0/).
+
+When there's both an SVG and PNG or GIF, the SVG is the source (easily editable later), but the PNG or GIF is derived and used for the MapLibre symbol layer.
```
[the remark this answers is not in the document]
dabreegster: Imagemagick convert did something weird. I manually fixed the white outside the circle. I think this is fine now, though setting icon-opacity to 0.5 for indicating existing TRs looks a little strange.

backend/src/movements.rs
```diff
@@ -20,13 +20,89 @@ impl MapModel {
...
+                if from.src_i == i.id {
+                    road_pointing_at_i.0.reverse();
+                }
+                // TODO Why +90 and +180? Kind of dunno, just making it look correct
+                let icon_angle =
+                    limit_angle(angle_of_pt_on_line(&road_pointing_at_i, pt.into()) + 90.0) + 180.0;
+
+                // Use the angle between the two roads to determine the type of turn
+                let bearing = euclidean_bearing(line.start, line.end);
+                let kind = classify_bearing(bearing);
```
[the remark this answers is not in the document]
dabreegster: Thank you for detangling this. I spent too many minutes with a notebook slowly losing my mind... part of my confusion was reasoning about how much to rotate the icons. It was very helpful to first just get the absolute bearing, then relative bearing to figure out turn type, then realize that rotating the icon is just the 1st absolute bearing.

I think this is all good now, from every case I've audited. @andrewphilipsmith very wisely made _two_ types of U turns back in v1 to distinguish left- from right-handed driving sides, and I've carried that forward here too.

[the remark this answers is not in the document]
dabreegster: Edit: the two u-turn variations are not about the driving side. Both kind exist in Kennington!
image
I'll revise that comment in the code

dabreegster:
OK, I've decided to reign in the scope of this PR and just get the icon visualization started, because it's complex enough. Editing can come later. Here's some pictures of how it looks:
image
image

There are known limits in the PR description -- particularly multiple overlapping icons. I will likely work on that next, or editing, or bits of both in parallel.

> re: moving the controls onto the map

Now a deferred decision for the next PR. :)   My only motivation was to uncrowd the left sidebar, yes. A vertical stack of commands could work too, and I guess even be similar to many paint program toolbars...

> we could organize our tools so that "draw many" becomes a context dependent sub-option of "Add Modal" rather than a top level item.

I like this idea too.

Honestly if we ditch the pico style and find little icons for some of these, stacking 4 or 5 horizontally in the sidebar will fit just fine too.

dabreegster:
Two more changes:

1) When you hover on a TR icon, no more arrow. Instead we just show the "from" and "to" road in their entirety. An arrow or something to distinguish the two could be nice, but along with the icon, it feels pretty clear. The old arrow outline style was less clear, and for U-turns especially when the icon was placed close to the end of the road, the arrow thickness was too much and line length too little to draw the arrowhead.

2) There's now simple handling of multiple TRs per origin road -- just stack them down the road. If we wind up with many TRs on a really short road, they're all going to smush into one point. I don't have a good approach for handling that, but also haven't organically come across an example yet.

[Screencast from 23-02-25 11:01:54.webm](https://github.com/user-attachments/assets/7e1adc31-13c5-40d1-9706-7adcaa5b03d1)

I'm now pretty happy with this PR as a first step. The open tasks in the PR description are more about interpreting TRs from OSM in the first place, not about visualization.

TFCx:
> 1. When you hover on a TR icon, no more arrow. Instead we just show the "from" and "to" road in their entirety. An arrow or something to distinguish the two could be nice, but along with the icon, it feels pretty clear. The old arrow outline style was less clear, and for U-turns especially when the icon was placed close to the end of the road, the arrow thickness was too much and line length too little to draw the arrowhead.

Could you make the segment a gradient from black to red ? or 2 segments (black until the junction and red after the junction) ?

(Damn, I know it's such a small modification, I should try to build the project locally and just do the test myself :/ I just don't have any time right now. :( )

backend/src/lib.rs
```diff
@@ -150,6 +150,11 @@ impl LTN {
         Ok(serde_json::to_string(&self.map.filters_to_gj()).map_err(err_to_js)?)
     }
 
+    #[wasm_bindgen(js_name = renderTurnRestrictions)]
+    pub fn render_turn_restrictions(&self) -> Result<String, JsValue> {
```
michaelkirk, line 154: Not a problem, but trying to understand: 

You use the term `render` in some of these methods - does it connote something in particular? (vs. say `get_turn_restrictions`)

web/src/layers/ModalFilterLayer.svelte
```diff
@@ -3,6 +3,8 @@
   import { emptyGeojson } from "svelte-utils/map";
   import { layerId } from "../common";
   import { backend, mutationCounter } from "../stores";
+  // TODO Maybe make another component wrapping both of these
```
michaelkirk, line 6: re: comment - "both"? What's the other layer? Or maybe this is todone?

michaelkirk (approved):
LGTM!

One minor request: 

I think it make sense for the turn restriction icons to follow the same visibility rules as the modal filters, addressed in #167 (minZoom  13), for the same reasons - it's not usable when zoomed out and looks chaotic.

Screenshot 2025-02-24 at 11 09 40

michaelkirk:
Yeah, the turn icons and the hover behavior make these really clear. Good work!

backend/src/lib.rs
dabreegster: Hmm, no. Looks like a habit I just retained from widgetry days. If the intent is only for visualization, I tend to use it. I wouldn't mind switching to `get` more consistently.

web/src/layers/ModalFilterLayer.svelte
dabreegster: Clarified. At first I was going to make every call-site include TurnRestrictionLayer, but it felt redundant, because it's all the same places as ModalFilterLayer. To be more pedantic, we could rename one of these or have some kind of `IconLayer`? But for now, I'm just punting on this

dabreegster:
Good point about the minzoom, thanks -- added!

> Could you make the segment a gradient from black to red ? or 2 segments (black until the junction and red after the junction)?

How does the user know which color is which? In my mind, drawing some kind of two-line arrow would be a little more clear. Happy to try this out later, I'll note it in #126. Thanks for the feedback/idea!

michaelkirk:
>> Could you make the segment a gradient from black to red ? or 2 segments (black until the junction and red after the junction)?
>
> How does the user know which color is which? In my mind, drawing some kind of two-line arrow would be a little more clear. Happy to try this out later, I'll note it in https://github.com/a-b-street/ltn/issues/126. Thanks for the feedback/idea!

By itself the road coloring would be ambiguous. But with the turn-restriction sign, I personally haven't found a case where it feels unclear in practice. Here's what it looks like in action:

https://github.com/user-attachments/assets/a3a1c730-dda0-48ed-aabb-49d8aac21ab0


Now that it's merged, you can give it a try and let us know what you think: https://a-b-street.github.io/ltn/

michaelkirk:
> re: moving the controls onto the map

In your work on turn restriction editing, feel free to propose something temporary if its expedient, and I can take another quick pass at it after the fact if it's convenient for you.

dabreegster:
> In your work on turn restriction editing, feel free to propose something temporary if its expedient, and I can take another quick pass at it after the fact if it's convenient for you.

I'm going to stick with the left sidebar and just change to icons with tooltips -- there are already some from the v1 tool

PR #150, opened 2025-02-19
Fix #149 for editing auto boundaries

### LINKED ISSUE #149: Ability to edit auto-generated boundaries
Currently the "Edit this boundary" feature is broken for autogenerated neighborhood boundaries, since there are no waypoints. (I only just now learned about waypoints!)

https://github.com/user-attachments/assets/eeb0a3cd-18e0-4d55-b77b-bc1ef667a69b

Note: we've also talked about having some high level way to "merge" adjacent auto-generated neighborhoods, but I think this is a separate issue.

A crude solution could be to create a "waypoint" for every vertex on the perimeter. 

A friendlier solution would then remove redundant waypoints as much as possible to simplify the shape without changing the resultant polygon.

dabreegster:
After fix:
image

See `invert_study_area_boundary` in Rust for how this feature is made in the first place.

It'd still be useful to reduce the waypoints to make this sanely editable. I've been meaning to write the same thing for NPW. Given the full desired geometry, come up with minimal waypoints that achieve the same thing.

### DIFF
--- a/web/src/SetBoundaryMode.svelte
+++ b/web/src/SetBoundaryMode.svelte
@@ -24,7 +24,9 @@
       // version of this tool, or an auto-generated boundary. Just
       // "backfill" by using the full geometry as freehand points.
       // Editing will be very painful in practice, but it won't break.
-      $waypoints = existing.geometry.coordinates[0].slice(1).map((point) => {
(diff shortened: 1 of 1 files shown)


michaelkirk:
Nice!

PR #152, opened 2025-02-20
Prioritize neighborhood by SIMD (UI only)

michaelkirk:
~~You don't have to review this yet, it's incomplete.~~ Go ahead and review!

Notably, the SIMD data I'm using is completely fake, but I'd like to demonstrate some of the UI tomorrow, and thought I'd share in case you wanted a preview.

Of course feel free to give feedback on the code too if you have any at this point.

**updated screencapture as of February 20, thursday eve. pacific time:**

https://github.com/user-attachments/assets/852668c9-e868-4436-b1e8-5a513bf7574a


web/src/PickNeighbourhoodMode.svelte
```diff
@@ -160,36 +168,88 @@
...
-            >
-              <img src={editIcon} alt="Rename neighbourhood" />
-            </button>
-            <button
-              class="secondary"
-              on:click={() => deleteNeighbourhood(name)}
-            >
-              <img src={deleteIcon} alt="Delete neighbourhood" />
-            </button>
+            <span class="neighbourhood-name">
```
michaelkirk, line 190: FYI - I'm making some pretty broad changes to this component, so try to avoid it for a bit if you can to minimize conflicts.

web/src/SetBoundaryMode.svelte
```diff
@@ -43,8 +43,9 @@
           mode: "neighbourhood",
         };
       } catch (err) {
+        console.log("error when setting hand drawn neighborhood", err);
         window.alert(
-          "Known georust bug hit, sorry. You may need to just refresh the page now.",
+          "Sorry, an error occurred. You may need to refresh the page for the application to continue working.",
```
michaelkirk, line 48: Stop tarnishing the brand! 🤣 

(truly - I think the old message is no longer relevant, so I made it more generic)

web/src/context/Population.svelte
```diff
@@ -1,23 +1,37 @@
...
+  let aggregateSIMDToNeighbourhoods = false;
   let showDensity = false;
 
-  // Color ramp from https://www.ons.gov.uk/census/maps/choropleth, dark to light.
-  let simdColorScale = ["#080C54", "#186290", "#1F9EB7", "#80C6A3", "#CDE594"];
-  // The percentiles are [1, 100]. The 5 colors cover 4 each.
-  let simdLimits = [0, 20, 40, 60, 80, 100];
-
+  $: neighbourhoods = emptyGeojson();
+  onMount(() => {
```
michaelkirk, line 21: I'm not sure why this needs to be so complicated, but given that I might delete it I didn't dig too deep yet.

web/src/context/Population.svelte
```diff
@@ -27,7 +41,10 @@
 {#if showSIMD}
   <SequentialLegend colorScale={simdColorScale} limits={simdLimits} />
   <p>Darker colours are more deprived</p>
-
+  <label>
+    <input type="checkbox" bind:checked={aggregateSIMDToNeighbourhoods} />
```
michaelkirk, line 45: This is sort of a UI experiment. I'm not sure it's actually useful.

web/src/wasm.ts
```diff
@@ -16,6 +17,11 @@ import type { Intersection, IntersectionFeature } from "./common/Intersection";
 // here aside from parsing and making the API nicer for both the Rust and TS
 // code. This is also a step towards moving to using web workers.
 
+export type NeighbourhoodBoundaryFeature = Feature<
```
michaelkirk, line 20: I like having these types available to use throughout the svelte app.

backend/src/map_model.rs
```diff
@@ -53,9 +54,100 @@ pub struct MapModel {
     // TODO Keep edits / state here or not?
     pub undo_stack: Vec<Command>,
     pub redo_queue: Vec<Command>,
-    // Stores boundary polygons in WGS84, with ALL of their GeoJSON props.
-    // TODO Reconsider
-    pub boundaries: BTreeMap<String, Feature>,
+    pub boundaries: BTreeMap<String, NeighbourhoodStats>,
```
michaelkirk, line 57: Note: I'm proposing we no longer tote around these opaque boundary `Feature`'s created on the frontend. 

Instead, that GeoJSON is converted to a NeighbourhoodDefinition struct at the js/wasm boundary, and we only deal with that on the rust side.

michaelkirk (commented):
Ok, I think this is ready for review. 

Note that it *only* covers the UI and data plumbing portion of the work, not actually packaging and using real SIMD data, since the methodology is still TBD.

So, it begs the question: are you OK to merge this with the "Fake SIMD" metric visible or should I hide that portion for now? It's easy enough to hide, but it might be helpful if any of the sustrans folks want to demo it.

Keep in mind: 
1. It's only visible in CNT projects
2. It's clearly labeled "Fake SIMD" - so I don't think anyone will get the wrong idea.

Screenshot 2025-02-20 at 19 03 20

michaelkirk:
Note: I updated the screen capture in the PR description if you want a preview.

michaelkirk:
BTW I'd recommend squashing all these commits upon merge. It was a bit of a round-a-bout effort.

web/src/AutoBoundariesMode.svelte
```diff
@@ -136,21 +164,68 @@
...
-        on:click={add}
+        layout={{
+          visibility: selectedPrioritization == "none" ? "visible" : "none",
+        }}
+        manageHoverState
         hoverCursor="pointer"
-      />
+        on:click={add}
+      >
+        <!--
```
dabreegster, line 176: https://github.com/dimfeld/svelte-maplibre/blob/master/src/lib/Popup.svelte
The new version using Svelte 5 fixes this, I think, by making a generic type for the component. (I don't think the fix depends on Svelte 5, it was just made during the migration.)  I haven't had a proper chance to play around with it, so I don't know how to specify the type parameter, but I agree something like this would be so awesome to have.

web/src/AutoBoundariesMode.svelte
```diff
@@ -136,21 +152,70 @@
...
+          <br />
+          <b>SIMD (percentile):</b>
+          {props.simd.toFixed(1)}
+        </Popup>
+      </FillLayer>
 
-      <LineLayer
-        {...layerId("auto-boundaries-severances")}
-        filter={isLine}
+      <FillLayer
```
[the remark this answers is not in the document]
dabreegster: The alternative is to have one FillLayer, and make the popup contents and line-color and everything else change based on `selectedPrioritization`. I'm not actually sure what's more efficient in maplibre -- to have many layers with more fixed paint expressions and toggle visibility, or to have one layer and change it frequently. Both cases use one `<GeoJSON>` source, so they might be similar.

This approach is fine to start. We could try the other approach (https://github.com/nptscot/npw/blob/2c3a691d296bfe005d86c1f4158c6e9952087d9e/web/src/layers/roads/ReferenceRoads.svelte#L154) later if it seems useful.

web/src/PickNeighbourhoodMode.svelte
```diff
@@ -1,27 +1,31 @@
...
   } from "./stores";
+  import type { NeighbourhoodBoundaryFeature } from "./wasm";
 
   // Note we do this to trigger a refresh when loading stuff
-  $: gj = $backend!.toSavefile();
-  $: boundaryNames = gj.features
-    .filter((f: Feature) => f.properties!.kind == "boundary")
-    .map((f: Feature) => f.properties!.name);
-  $: edits = countEdits(gj);
+  $: neighbourhoods = $backend!.getAllNeighbourhoods();
```
dabreegster, line 23: Much cleaner :)

web/src/PickNeighbourhoodMode.svelte
```diff
@@ -159,20 +163,30 @@
...
-      </Link>
-    </div>
-
-    <ul>
-      {#each boundaryNames as name}
-        <li>
+    <h3 style="padding: 4px 8px; margin-bottom: 0">Neighbourhoods</h3>
+    <ul class="neighbourhood-list">
+      {#each neighbourhoods.features as { properties: { name, area_km2, simd } }}
+        <!--
```
dabreegster, line 169: Hmm. I think `JoinedData` can also work with GJ, by leaving this unset.

```
  export let sourceLayer: string | undefined = undefined;
```

But I haven't played with it here. The other brute force alternative is to make a second <GeoJSON> with zero or one features, and set it here. That's messier, requires copying the feature over.

I'll have a go at this later.

web/src/PickNeighbourhoodMode.svelte
```diff
@@ -214,24 +262,103 @@
...
+          pickNeighbourhood(notNull(e.detail.features[0].properties).name)}
       >
         <Popup openOn="hover" let:props>
-          <p>{props.name}</p>
+          <h2>{props.name}</h2>
+          <b>SIMD:</b>
+          {props.simd.toFixed(1)}
+        </Popup>
+      </FillLayer>
+      <FillLayer
```
dabreegster, line 309: Hmm, now that I see how the multiple fill layers are duplicated here too, I think one Svelte component that's responsible for one FillLayer and takes in `selectedPrioritization` as a prop is sounding nicer. But we can leave as an idea for later.

web/src/prioritization/PrioritizationSelect.svelte
```diff
@@ -0,0 +1,49 @@
+<script lang="ts">
+  import { SequentialLegend } from "svelte-utils";
+  import { areaLimits, simdColorScale, simdLimits } from "../common/colors";
+  import { projectName } from "../stores";
+
+  export let selectedPrioritization: "none" | "area" | "simd";
+
+  function setSelectedPrioritizationFromURL() {
```
dabreegster, line 8: Definitely useful. I've been uncertain how much to make the URL reflect the current app state and be more easily shareable. There's so much state everywhere, and lots of it depends on having an exact project file or something. Are you imagining we do more of this?

web/src/AutoBoundariesMode.svelte
```diff
@@ -116,10 +135,19 @@
 
   <div slot="map">
     <GeoJSON data={gj} generateId>
+      <LineLayer
+        {...layerId("auto-boundaries-severances")}
+        filter={isLine}
+        manageHoverState
+        paint={{
+          "line-color": hoverStateFilter("black", "red"),
```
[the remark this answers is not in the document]
michaelkirk: I think you know this, but to be clear, this layer was pre-existing, it just got moved around a bit in the diff.

web/src/prioritization/PrioritizationSelect.svelte
michaelkirk: > There's so much state everywhere, and lots of it depends on having an exact project file or something. 

Yeah, that's a good point, and something to be aware of. In this case, which prioritization metrics are valid would be universal to all CNT users (though irrelevant for non-cnt users).

To your point, I can imagine specific url params that would be meaningless across projects (e.g. `?intersectionId=4`)

> Are you imagining we do more of this?

Yes! Though I will confess, my immediate concern could have been solved without this. I was just trying to storing the state so that when you navigate around the field doesn't keep resetting. Another way to achieve that would be to use localStorage.  

But thinking long term, I think it would be good to build towards making this work more like a "normal" web app because it's a useful pattern that people expect.

Broadly, there's two things I'd like to encode in the URL:
1. what page/mode you're in
2. parameters that apply to that page (like the selectedPrioritization)

As an example, I think you could understand where this URL should take you:
https://cnt.ltn.org/projects/my-cnt-project/neighbourhoods/new-from-autogenerated-boundary?selectedPrioritization=simd

So what constitutes a "page" in our app? I think anything that currently ends up in the "breadcrumb" navigation would be a good start - each of those should have a unique URL, and that "mode" should be reachable directly from that URL without having to re-click through.

To complicate this vision: because all the state is local, having readable url's like https://cnt.ltn.org/projects/my-project/neighbourhoods/my-dundee might give people the *wrong* idea that they can just send that URL to a friend and their friend will be able to see the project.

Since all project data is currently local only, a URL like that will only work for the creator of the neighbourhood. We'd probably need  to facilitate this case by showing, in lieue of an opaque 404, something to the effect of "It looks like you're trying to access a project that you don't  have yet - ask the creator to send you the exported project and then [upload here]"

I think longer term the idea actually **is** to support sharing urls like this and some kind of server-side application storage, right?

So anyway, back to where we are today, we don't currently have a router at our disposal (right?). "sveltekit" has some kind of a router builtin, but I don't know how easy it is to just "drop it in" to our app - my impression is that sveltekit apps require a pretty specific layout, which could be a pretty big change. I believe there are also "stand alone" routers that we could pursue which might be less disruptive to integrate, or we could build something custom.

So, again this is mostly high minded rambling at this point. I don't think this is on the critical path in the short term, but I think it'd be a great direction to move towards, especially if we could figure out a way to do it that's not disruptive.

dabreegster: > I think longer term the idea actually is to support sharing urls like this and some kind of server-side application storage, right?

Yes, at which point this "deep-app-state" URL sharing becomes really really compelling.

> "sveltekit" has some kind of a router built in, but I don't know how easy it is to just "drop it in" to our app - my impression is that sveltekit apps require a pretty specific layout, which could be a pretty big change

Correct, I'd be quite hesitant to do this soon, unless sharing URLs becomes important. I haven't played with other off-the-shelf routers, but I've done something custom like this before:

https://github.com/acteng/atip/blob/main/src/lib/browse/layers/url.ts
Example use: https://github.com/acteng/atip/blob/02b725b9103fc160cc8a357e2c1c87bc6bab285d/src/lib/browse/layers/points/Stats19.svelte#L24
It's quite verbose to writer parsers/stringifers. Presumably the real router libs make this nicer.

Besides the problem with sharing files between people, I could also imagine old URLs / pressing back would sometimes break. If you rename/delete a neighbourhood boundary (in the world where we have multiple), old URLs become stale.

No concrete ideas right now to make this work, but it's a good thing to think about...

PR #160, opened 2025-02-20
Cleanup help button and action buttons on project list and neighbourhood list

### LINKED ISSUE #151: For button icons, use icon font rather than SVG images
A small UI/UX thing... 

currently our button icons are SVG.

Using an icon font (something like https://lucide.dev/icons or https://fonts.google.com/icons) would allow us to apply different style treatments to the icons depending on context.

Eg. while polishing up the "project list" I wanted to mark the destructive actions as red:

Image

Ideally I'd change the trash icon's stroke to be white for better contrast, but I'm not aware of a nice way to do this without editing the svg file. And I maybe don't want to change it in *all* contexts (I could make a separate trash-white.svg file, but that's a little annoying too). 

With an icon font, we can apply a css rule to scope the styling. 

Icon fonts are a common thing and easy to add, but I wasn't sure if you were opposed from an idealogical standpoint. It's not a big deal to me if we don't do this.

michaelkirk:
This introduces the "lucid" icon font, which apparently just renders SVGs, but they've devised a nice API so that you can tweak the coloring (and stroke width).

I guess one advantage over a traditional icon **font**, is that using our build system (vite) it's efficient about which assets get included in the compiled bundle.

I realize we might not end up using either of these screens for CNT, but we will use the nav and help buttons which I spruced up. I didn't spend a lot of time on this, and it's nice to clean it up in the interim, and for non-CNT use. 

Fixes #151 

**before:**

before_projects

before_neighbourhoods

**after:**

after_projects

after_neighbourhoods


dabreegster (approved):
Yeah, this looks fantastic! Thanks for fighting through the clashes with pico.

PR #161, opened 2025-02-21
Make the *current* breadcrumb stand out

michaelkirk:
Just a cute little UI flourish that helps establish where you *are* in the breadcrumb. This might be especially effective for red/green color blind people like me for which the current "link" colored text does little to differentiate.

https://github.com/user-attachments/assets/44f9a0c8-9f70-43d7-96a9-6b94f0919444

### DIFF
--- a/web/src/App.svelte
+++ b/web/src/App.svelte
@@ -264,4 +264,9 @@
     margin-right: -32px;
     margin-left: -16px;
   }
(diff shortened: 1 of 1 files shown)


dabreegster (approved):
I almost want to suggest this upstream to pico, if the main repo wasn't now unmaintained ^^;   Part of the reason I wanted to use Pico in the first place was to assume they'd make reasonable defaults for things like this. Thanks for the fix!

Issue #145, opened 2025-02-15
Confusing artifacts in autogenerated boundaries


michaelkirk:
I know that getting autogenerated boundaries perfect is not going to happen.

One problem in particular that we could make progress on: the existence of tiny neighborhoods. 

A couple heuristics that might help: 

1. Any neighborhood contained by another neighborhoods exterior should be merged into that neighborhood:

<img width="291" alt="Image" src="https://github.com/user-attachments/assets/0e335c62-1e42-4bbf-b372-302e35e3440e" />

2. Any neighborhood under a certain size should be merged into it's neighbor. We could probably do something smart to try to pick a good candidate neighbor, but even picking randomly might be a substantial improvement to the status quo.

(from glasgow city)

dabreegster:
In your your first example, that's one big green neighbourhood. The black lines in the middle are strips of railway. They're just showing the severances used to split things. If we just stopped drawing those and made the boundary outlines thicker, that might be a better view to show.

A related case to 1 is a boundary with a hole in it for water, like 

![Image](https://github.com/user-attachments/assets/adc5414b-2ba6-48e0-bf3c-7068d2afc4f8) in Liverpool at #14.32/53.41006/-3.06012

michaelkirk:
Mostly solved by #162 - I think the only remaining work is to do something about the lakes, which produce holes in the boundaries.

dabreegster:
> Do you think this could be as simple as removing all holes? Or are there some holes we want to keep?

I haven't everywhere, but I bet we could just remove holes. Only counterexample I can imagine is a circular highway on some kind of island, itself surrounded by big roads. But if that inner island's road network has a big-road bridge connecting to the rest, it won't wind up as a hole.

PR #162, opened 2025-02-21
Stop displaying severances in the auto boundaries mode. #145

dabreegster:
Before, severances are thick black lines that turn red when you hover on them. This was more intended for debug/development; they're confusing to the end user.
image

After, we just show a thin outline around the clickable boundaries:
image

The holes in some boundaries due to ponds still looks odd, but that's something we can fix later.

### DIFF
--- a/backend/src/auto_boundaries.rs
+++ b/backend/src/auto_boundaries.rs
@@ -28,29 +28,16 @@ impl MapModel {
                     "tertiary_link",
                 ],
             ) {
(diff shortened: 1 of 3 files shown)


michaelkirk (approved):
Nice!

michaelkirk:
> The holes in some boundaries due to ponds still looks odd, but that's something we can fix later.

Do you think this could be as simple as removing *all* holes? Or are there some holes we want to keep?

PR #163, opened 2025-02-21
Highlight the neighbourhood on the map when hovering on an item in the sidebar

dabreegster:
[Screencast from 21-02-25 10:13:47.webm](https://github.com/user-attachments/assets/a7822fea-d14a-47cc-bfa2-642d8555fdf4)

Also refactor to one FillLayer

### DIFF
--- a/web/src/PickNeighbourhoodMode.svelte
+++ b/web/src/PickNeighbourhoodMode.svelte
@@ -112,14 +112,11 @@
   function fillColor(
     selectedPrioritization: "none" | "area" | "simd",
   ): DataDrivenPropertyValueSpecification<string> {
-    let color: DataDrivenPropertyValueSpecification<string> = "red";
-    if (selectedPrioritization == "none") {
-      color = "black";
-    } else if (selectedPrioritization == "simd") {
-      color = makeRamp(["get", "simd"], simdLimits, simdColorScale);
-    } else if (selectedPrioritization == "area") {
-      color = makeRamp(["get", "area_km2"], areaLimits, simdColorScale);
-    }
+    let color = {
+      none: "black",
+      simd: makeRamp(["get", "simd"], simdLimits, simdColorScale),
+      area: makeRamp(["get", "area_km2"], areaLimits, simdColorScale),
+    }[selectedPrioritization];
     return [
       "case",
       ["==", ["feature-state", "highlight"], "yes"],
--- a/web/src/PickNeighbourhoodMode.svelte
+++ b/web/src/PickNeighbourhoodMode.svelte
@@ -1,7 +1,12 @@
 <script lang="ts">
   import type { FeatureCollection } from "geojson";
   import { CirclePlus, Pencil, Trash2 } from "lucide-svelte";
-  import { FillLayer, GeoJSON, hoverStateFilter } from "svelte-maplibre";
+  import {
+    FillLayer,
+    GeoJSON,
+    hoverStateFilter,
+    JoinedData,
+  } from "svelte-maplibre";
   import { downloadGeneratedFile, notNull } from "svelte-utils";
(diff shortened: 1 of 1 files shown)


web/src/PickNeighbourhoodMode.svelte
```diff
@@ -103,7 +109,35 @@
...
+    } else if (selectedPrioritization == "simd") {
+      color = makeRamp(["get", "simd"], simdLimits, simdColorScale);
+    } else if (selectedPrioritization == "area") {
+      color = makeRamp(["get", "area_km2"], areaLimits, simdColorScale);
+    }
+    return [
+      "case",
+      ["==", ["feature-state", "highlight"], "yes"],
+      "yellow",
+      // @ts-expect-error MapLibre expression types are weird, but this really does work
```
dabreegster, line 127: Another use of moving this into the <script> section is being able to do this type cast / TS override. We can't easily do it in the section below in Svelte 4.

web/src/PickNeighbourhoodMode.svelte
```diff
@@ -262,62 +286,20 @@
   </div>
 
   <div slot="map">
-    <GeoJSON data={neighbourhoods} generateId>
+    <GeoJSON data={neighbourhoods} promoteId="name">
```
dabreegster, line 289: Honestly `JoinedData` could use better docs or more examples. We need to first tell maplibre to use the `name` property as the unique ID, so the feature selector for `setFeatureState` works. Then this all works out.

web/src/PickNeighbourhoodMode.svelte
```diff
@@ -262,62 +286,20 @@
...
         filter={["==", ["get", "kind"], "boundary"]}
         paint={{
-          "fill-color": "black",
-          "fill-opacity": hoverStateFilter(0.3, 0.5),
-        }}
-        layout={{
-          visibility: selectedPrioritization == "none" ? "visible" : "none",
-        }}
-        manageHoverState
-        bind:hovered={hoveredMapFeature}
```
dabreegster, line 277: I didn't realize this till now, but it's actually a bit scary to previously have multiple FillLayers all binding to the same variable. It worked fine before because only one layer was visible at a time, but it feels confusing -- 3 simultaneous writers. I think using just one layer results in simpler code, what do you think?

web/src/PickNeighbourhoodMode.svelte
```diff
@@ -327,11 +309,18 @@
...
         <Popup openOn="hover" let:props>
           <h2>{props.name}</h2>
-          <b>Area:</b>
-          {props.area_km2.toFixed(1)} km²
+
+          {#if selectedPrioritization == "none" || selectedPrioritization == "area"}
+            <b>Area:</b>
+            {props.area_km2.toFixed(1)} km²
+          {:else if selectedPrioritization == "simd"}
+            <b>Fake SIMD:</b>
```
dabreegster, line 317: Nit: to really make sure nobody gets confused in the next few days, labeling this fake here too

dabreegster (commented):
Argh, my screencasts don't capture the cursor. I promise this works, or go try it yourself

web/src/PickNeighbourhoodMode.svelte
michaelkirk: I justified 3 simultaneous writers because they were all displayed mutex, but I think what you've done here is inarguably an improvement!

web/src/PickNeighbourhoodMode.svelte
michaelkirk: I did figure out `JoinedData.idCol`, but I missed the `GeoJSON.promoteId` - thanks for figuring it out!

michaelkirk (approved):
Wonderful!

PR #164, opened 2025-02-21
Add a stats19 contextual layer.

dabreegster:
#106 

Most code and styling decisions are lifted from https://github.com/acteng/atip-data-prep/ and https://github.com/acteng/atip.
image

### DIFF
--- a/data_prep/scotland/get_reference_layers.sh
+++ b/data_prep/scotland/get_reference_layers.sh
@@ -80,6 +80,39 @@ function population {
   tippecanoe tmp/population.geojson -zg --generate-ids -l population -o $OUT/population.pmtiles
 }
 
+function stats19 {
+  # Use code from https://github.com/acteng/atip-data-prep/tree/main/stats19
+  git submodule init
+  git submodule update
+
+  cd atip-data-prep/stats19
+  mkdir -p input tmp
+
+  # URLs from https://www.data.gov.uk/dataset/cb7ae6f0-4be6-4935-9277-47e5ce24a11f/road-safety-data
+  if [ ! -e "input/dft-road-casualty-statistics-casualty-1979-latest-published-year.csv" ]; then
+    download_to_subdir input https://data.dft.gov.uk/road-accidents-safety-data/dft-road-casualty-statistics-casualty-1979-latest-published-year.csv
+  fi
+  if [ ! -e "input/dft-road-casualty-statistics-collision-1979-latest-published-year.csv" ]; then
+    download_to_subdir input https://data.dft.gov.uk/road-accidents-safety-data/dft-road-casualty-statistics-collision-1979-latest-published-year.csv
+  fi
+
+  cargo run --release
+
+  download_to_subdir input https://raw.githubusercontent.com/georgique/world-geojson/refs/heads/develop/areas/united_kingdom/scotland.json
+  mapshaper tmp/stats19.geojson -clip input/scotland.json -o tmp/stats19_clipped.geojson
+
+  tippecanoe tmp/stats19_clipped.geojson \
+	  --force \
+	  --generate-ids \
+	  -l stats19 \
+	  -zg \
+	  --drop-densest-as-needed \
+	  --extend-zooms-if-still-dropping \
+	  -o ../../$OUT/stats19.pmtiles
+
+  cd ../../
+}
+
 function get_scotland_osm {
   # Download Scotland OSM data
   if [ ! -f tmp/scotland-latest.osm.pbf ]; then
@@ -104,6 +137,7 @@ download_to_subdir() {
 #railway_stations
 #bus_routes
 #population
+stats19
 
 echo "For maintainer only:"
 echo "  mv $OUT/* ~/cloudflare_sync/cnt_layers/"
--- a/data_prep/scotland/get_reference_layers.sh
+++ b/data_prep/scotland/get_reference_layers.sh
@@ -89,12 +89,8 @@ function stats19 {
   mkdir -p input tmp
 
   # URLs from https://www.data.gov.uk/dataset/cb7ae6f0-4be6-4935-9277-47e5ce24a11f/road-safety-data
-  if [ ! -e "input/dft-road-casualty-statistics-casualty-1979-latest-published-year.csv" ]; then
-    download_to_subdir input https://data.dft.gov.uk/road-accidents-safety-data/dft-road-casualty-statistics-casualty-1979-latest-published-year.csv
-  fi
-  if [ ! -e "input/dft-road-casualty-statistics-collision-1979-latest-published-year.csv" ]; then
-    download_to_subdir input https://data.dft.gov.uk/road-accidents-safety-data/dft-road-casualty-statistics-collision-1979-latest-published-year.csv
-  fi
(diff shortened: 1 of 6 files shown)


data_prep/scotland/get_reference_layers.sh
```diff
@@ -80,6 +80,39 @@ function population {
...
+    download_to_subdir input https://data.dft.gov.uk/road-accidents-safety-data/dft-road-casualty-statistics-casualty-1979-latest-published-year.csv
+  fi
+  if [ ! -e "input/dft-road-casualty-statistics-collision-1979-latest-published-year.csv" ]; then
+    download_to_subdir input https://data.dft.gov.uk/road-accidents-safety-data/dft-road-casualty-statistics-collision-1979-latest-published-year.csv
+  fi
+
+  cargo run --release
+
+  download_to_subdir input https://raw.githubusercontent.com/georgique/world-geojson/refs/heads/develop/areas/united_kingdom/scotland.json
+  mapshaper tmp/stats19.geojson -clip input/scotland.json -o tmp/stats19_clipped.geojson
```
dabreegster, line 102: With all of the layers in here, I'm imagining to use them for prioritization. we'd also convert to a gpkg or fgb or gmd or other format too.

web/src/context/ContextualLayers.svelte
```diff
@@ -15,7 +16,7 @@
   <div
     class="pico"
     style:background="grey"
-    style="display: flex; flex-direction: column"
+    style="display: flex; flex-direction: column; max-height: 80vh; overflow: auto"
```
dabreegster, line 19: Ran out of vertical space. The NPW UI mocks have some ideas -- I'd be happy to implement any ideas you have for short-term improvements, but I'm still a bit holding out to borrow ideas from there.

web/src/context/Stats19.svelte
```diff
@@ -0,0 +1,234 @@
...
+    pedestrians: true,
+    cyclists: true,
+    minYear: 2017,
+    maxYear: 2023,
+  };
+
+  function makeFilter(_: any): ExpressionSpecification {
+    let includeTypes: ExpressionSpecification = ["any"];
+    if (state.pedestrians) {
+      includeTypes.push(["get", "pedestrian"]);
```
dabreegster, line 24: The data we produce also has `horse_rider` and `other`, leftovers from what's needed for the original use case. When we start using this to match to neighbourhoods and count, we will also need to exclude. Maybe we should exclude upfront and be simpler...

data_prep/scotland/get_reference_layers.sh
```diff
@@ -80,6 +80,39 @@ function population {
   tippecanoe tmp/population.geojson -zg --generate-ids -l population -o $OUT/population.pmtiles
 }
 
+function stats19 {
+  # Use code from https://github.com/acteng/atip-data-prep/tree/main/stats19
```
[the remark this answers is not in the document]
michaelkirk: Submodules are annoying, but they solve a problem. I'm totally fine with it.

I do have this handy fix-all alias that I make a lot of use of:

`gsu='git submodule update --init --recursive'`

data_prep/scotland/get_reference_layers.sh
```diff
@@ -80,6 +80,39 @@ function population {
...
+  git submodule init
+  git submodule update
+
+  cd atip-data-prep/stats19
+  mkdir -p input tmp
+
+  # URLs from https://www.data.gov.uk/dataset/cb7ae6f0-4be6-4935-9277-47e5ce24a11f/road-safety-data
+  if [ ! -e "input/dft-road-casualty-statistics-casualty-1979-latest-published-year.csv" ]; then
```
michaelkirk, line 92: No biggie, but you can probably skip the -e check.

In case you missed it, `download_to_subdir` uses `--timstamping` option, so it won't actually download the file contents unless the server reports it as modified vs. your local versions timestamps.

data_prep/scotland/get_reference_layers.sh
michaelkirk: I'm imagining another entry into `data_prep/src/bin/generate-prioritization-layers` (name TBD) that does this.

I'm actually planning to get started on this today with the SIMD dataset. We won't process it yet, but I think in any case we're pretty confident we'll want all the population "zone" polygons in a format accessible to rust and they'll need to be pre-sliced by LAD and REGION... right?

web/src/context/Stats19.svelte
michaelkirk: Exclude as part of the data_prep process you mean? That seems reasonable.

web/src/context/Stats19.svelte
```diff
@@ -0,0 +1,234 @@
...
+      >
+        DfT stats19
+      </a>
+      dataset, as of 30 September 2024. Please note these limitations:
+    </p>
+    <ul>
+      <li>Only collisions between 2017 and 2023 are included</li>
+      <li>
+        This tool is intended to be used at high-zoom levels, while inspecting a
+        scheme or development area. Not all points are shown at lower zoom
```
michaelkirk, line 115: Take it or leave it copy nit: Do people know which zoom levels are high vs low? 

Personally I think "zoomed out" and "zoomed in" are more intuitive, but are kind of vague in their own way.

web/src/context/Stats19.svelte
```diff
@@ -0,0 +1,234 @@
...
+  let seriousColor = "#1F9EB7";
+  let slightColor = "#CDE594";
+  let severityLegend = {
+    Fatal: fatalColor,
+    Serious: seriousColor,
+    Slight: slightColor,
+  };
+</script>
+
+<button class="secondary" on:click={() => (show = !show)}>Stats19</button>
```
michaelkirk, line 99: Take it or leave it: How universal is "Stats19" as a concept? Can we call this  `Collisions` or `Collisions (stats19)` - it seems more likely to be meaningful to an avg person.

michaelkirk (approved):
Looks great! Other than this, do you have any imminent work on the context controls? I'd like to do a quick pass at fixing up the layout soon.

data_prep/scotland/get_reference_layers.sh
dabreegster: Sounds right to me, yep. I think starting this soon and starting to match these to generated nhood boundaries is a good idea, even before the scoring is nailed down. I can almost guarantee it's something we'll iterate on and tune for all cases anyway.

web/src/context/Stats19.svelte
dabreegster: Low energy now, but I'll take it as a TODO.

data_prep/scotland/get_reference_layers.sh
dabreegster: Oh nice! Done

web/src/context/Stats19.svelte
dabreegster: Done

web/src/context/Stats19.svelte
dabreegster: Great point, done

dabreegster:
> do you have any imminent work on the context controls? I'd like to do a quick pass at fixing up the layout soon.

Nope, go for it!

PR #165, opened 2025-02-21
shared workspace for rust deps

### LINKED ISSUE #110: Set up cargo workspaces
The `backend` and `data_prep` crates both do some crates.io overrides. A workspace would be easier to maintain.

michaelkirk:
This is especially helpful when doing patch over-rides, since `data_prep/scotland` depends on `../../backend`.

Fixes #110


dabreegster (approved):
Thank you for this cleanup!

PR #166, opened 2025-02-22
prep data for simd prioritization, and tie it into the app.

michaelkirk:
The current methodology uses area proportional accumulation, but we can revisit this when we hear back from Cici.

In the video I toggle the underlying simd data zones context on/off a few times. If you let your eyes blur, hopefully you can see how the aggregation in our generated boundaries is an echo of the underlying data zones.

https://github.com/user-attachments/assets/a2d9757c-3e46-4b67-bd02-9759e1d05ecf

Runtime to prepare the data is about 4 minutes on my machine, about 2/3 of which is spent
computing the tricky topology of the Highlands and Islands region.

> Time since start 244.382083083s

The autogen boundary page is noticeably slower, especially as you get to bigger areas.  But there is some low hanging fruit that I hope is sufficient to get things reasonably snappy.

We can merge now if you're OK with it, or I'm happy to hold off until some of the low hanging perf fruit has been picked early next week if you prefer.


backend/src/create.rs
```diff
@@ -133,12 +134,21 @@ pub fn create_from_osm(
     boundary_wgs84: MultiPolygon,
     study_area_name: Option<String>,
     demand: Option<DemandModel>,
+    mut population_zones_wgs84: Option<Vec<PopulationZone>>,
```
michaelkirk, line 137: The population_zone bytes are plumbed through a lot of layers. 

I'm kind of assuming once we add other metrics that we'll maybe consolidate some of the datasets, so that hopefully we're not plumbing through 5 or 6 different parameters through the entire app.

data_prep/scotland/src/bin/generate_prioritization.rs
```diff
@@ -0,0 +1,96 @@
...
+    // "population": 894,
+    pub population: u32,
+    // "area": 4388802.1221970674
+    // (unused - though maybe we would find it helpful for pre-computing density or to save the cost of calculating area live)
+}
+
+impl PopulationZoneInput {
+    fn read_all_from_file() -> Result<Vec<Self>> {
+        let population_zones = geojson::de::deserialize_feature_collection_str_to_vec(
+            // REVIEW: I think it's OK to just explode with useful error messages for the data_prep stuff, right?
```
dabreegster, line 77: Yep, this is just offline scripts you or I are running. Error recovery doesn't make sense

backend/src/boundary_stats.rs
```diff
@@ -9,12 +8,52 @@ pub struct BoundaryStats {
...
+        // area proportional accumulation:
+        // If  3/4 the boundary is in a zone with simd 100:   0.75 * 100 = 75
+        // and 1/4 the boundary is in a zone with  simd 40:  +0.25 *  40 = 10
+        //                  than the computed simd will be:          sum = 85
+        // TODO: optimize this - populationZone should at least be prepared geometries, and
+        // TODO: maybe also introduce an overall RTree to get intersection candidates quickly
+        let mut simd = 0.0;
+        if let Some(population_zones) = population_zones {
+            for population_zone in population_zones {
+                if population_zone.geometry.intersects(polygon) {
```
dabreegster, line 25: I don't know much about the implementations of these -- is there lots of "wasted work" to first check for intersections, then calculate exactly what it is? Presumably the first is much faster, and something like a bounding box overlap check isn't a useful alternate first pass?

backend/src/boundary_stats.rs
```diff
@@ -9,12 +8,52 @@ pub struct BoundaryStats {
...
+    // "imd_rank": 4691,
+    // (unused)
+
+    // "imd_percentile": 68,
+    pub imd_percentile: u8,
+
+    // "population": 894,
+    pub population: u32,
+    // "area": 4388802.1221970674
+    // (unused - though maybe we would find it helpful for pre-computing density or to save the cost of calculating area live)
```
dabreegster, line 58: These notes are very useful for development/code review, but as we iterate on this next week with methodology, I'd eventually want to clean these up for the more-final code

backend/src/create.rs
dabreegster: Oh, definitely. Maybe there's one optional demand file per study area, and one optional "everything else" file per study area, and just those two are plumbed around. I was assuming all of the prioritization data gets bundled into one blob.

dabreegster (approved):
This is a great start! I'll upload the new files.

One thing I think we want to do next is rename "population zone" to something more general like "prioritization data" or even just "contextual data" or something. If we wind up with 10 metrics used, I don't think there should be 10 separate files loaded and plumbed around, but one struct bundle. That should reduce boilerplate and also latency for many files.

There's one unit test that needs an update; I'll push a commit to this branch and merge for now

> Runtime to prepare the data is about 4 minutes on my machine, about 2/3 of which is spent computing the tricky topology of the Highlands and Islands region.

... Huh. Entire cargo run is 17s for me. Total size of the prioritization output dir is 122MB. I'm just going to assume this comment was a little old, and you switched to prepared geometry after writing it...

michaelkirk:
>> Runtime to prepare the data is about 4 minutes on my machine, about 2/3 of which is spent computing the tricky topology of the Highlands and Islands region.
>
>... Huh. Entire cargo run is 17s for me. Total size of the prioritization output dir is 122MB. I'm just going to assume this comment was a little old, and you switched to prepared geometry after writing it...

I forgot `--release` 🤦 

It was actually much much slower  without the prepared geometries.

backend/src/boundary_stats.rs
michaelkirk: Will do. In case it wasn't obvious these comments were intended to document the other fields available in the input, including the ones we're not yet using.

Issue #78, opened 2025-01-12
Pick neighbourhood page hard to use


dabreegster:
At low zoom with many filters, the neighbourhood boundaries can be hidden!!
![image](https://github.com/user-attachments/assets/20b504b5-3306-459b-a604-f5dde9907f26)
The boundary is there somewhere:
![image](https://github.com/user-attachments/assets/d2e0de4b-af8e-4a7a-8aee-d605f6afbfcb)

What should happen design-wise on this page?

XioNoX:
I usually create projects about the size of the neighborhood I work on, so this issue doesn't happen to me.

dabreegster:
Thanks, that's good to know. In some of our other use cases, the map will already be precreated and quite large, so this still will be something to figure out.

I think part of the problem is too many filters shown when unzoomed. In #6 I'm realizing some places have lots of OSM barrier nodes on a dead-end street. Those can't impact through-traffic, so it feels like just noise to show them.

michaelkirk:
One easy thing we could do is set the min-zoom. 

How would you feel about starting with that  and seeing if it's still enough of a nuisance to warrant a more challenging solution?

dabreegster:
Min zoom sounds like a great start, yes. I also want to try some work to detect existing filters as "important" or not based on deadends. It might be a heuristic mess hard to get right, but maybe we could have three zoom levels showing no icons, only important ones (maybe even just user-edited?), and all

PR #167, opened 2025-02-22
Hide filters when zoomed out - it's too cluttered.

michaelkirk:
FIX https://github.com/a-b-street/ltn/issues/78

There's probably something fancier we could do, but this was easy and solves the worst of it.

**before:**

https://github.com/user-attachments/assets/6347c468-8ade-438f-b652-32a433fe2798

**after:**

https://github.com/user-attachments/assets/5a8f0f80-dd83-49ee-a96e-7f5e516ad04f


dabreegster (approved):
Yeah, why not. This is simple, and there's nothing interesting communicated from far away

PR #168, opened 2025-02-22
Speed up boundary stats

michaelkirk:
Followup to https://github.com/a-b-street/ltn/pull/166#discussion_r1966478092, to speed up the calculation using a prepared geometry.

```
build stats: LAD_Dundee City
                        time:   [1.8475 ms 1.8544 ms 1.8629 ms]
                        change: [-29.968% -29.750% -29.471%] (p = 0.00 < 0.05)
                        Performance has improved.
```

Feel free to squash these commits (or at least the Temporary/Revert ones)

### DIFF
--- a/backend/src/boundary_stats.rs
+++ b/backend/src/boundary_stats.rs
@@ -22,8 +22,8 @@ impl BoundaryStats {
         let mut simd = 0.0;
         if let Some(population_zones) = population_zones {
             for population_zone in population_zones {
-                if population_zone.geometry.intersects(polygon) {
-                    let overlap = polygon.intersection(&population_zone.geometry);
+                let overlap = polygon.intersection(&population_zone.geometry);
+                if overlap.unsigned_area() > 0.0 {
                     let overlap_ratio = overlap.unsigned_area() / area_meters;
                     simd += overlap_ratio * population_zone.imd_percentile as f64
                 }
--- a/backend/src/boundary_stats.rs
+++ b/backend/src/boundary_stats.rs
@@ -22,8 +22,8 @@ impl BoundaryStats {
         let mut simd = 0.0;
         if let Some(population_zones) = population_zones {
             for population_zone in population_zones {
-                let overlap = polygon.intersection(&population_zone.geometry);
-                if overlap.unsigned_area() > 0.0 {
+                if population_zone.geometry.intersects(polygon) {
+                    let overlap = polygon.intersection(&population_zone.geometry);
                     let overlap_ratio = overlap.unsigned_area() / area_meters;
                     simd += overlap_ratio * population_zone.imd_percentile as f64
                 }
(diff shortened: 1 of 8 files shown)


backend/src/boundary_stats.rs
```diff
@@ -22,10 +22,14 @@ impl BoundaryStats {
         let mut simd = 0.0;
         if let Some(population_zones) = population_zones {
             for population_zone in population_zones {
-                if population_zone.geometry.intersects(polygon) {
-                    let overlap = polygon.intersection(&population_zone.geometry);
```
michaelkirk, line 26: re: https://github.com/a-b-street/ltn/pull/166#discussion_r1966478092

> is there lots of "wasted work" to first check for intersections, then calculate exactly what it is?

`intersects` is typically a little faster than `intersection`, so it's nice to do the quick path first.

I measured the difference in a temporary commit in the PR: https://github.com/a-b-street/ltn/pull/168/commits/29925abf75ac446b2aae6d1a7577b09511f8c13a

But more importantly, we can use a prepared geometry for `intersects` to speed it up even more (but we cannot do this with `intersection`).

For a little explanation, `intersects` is some topology code internal to geo, `intersection` is a wrapper around i_overlay.

backend/src/test_fixtures.rs
```diff
@@ -35,36 +45,83 @@ impl NeighbourhoodFixture {
...
-        let pbf_path = format!("../web/public/severance_pbfs/{study_area_name}.pbf");
-        let input_bytes = std::fs::read(&pbf_path)?;
-
-        let boundary_path = format!("../web/public/boundaries/{study_area_name}.geojson");
-        let boundary: Feature = std::fs::read_to_string(&boundary_path)?.parse()?;
-        // All test study area boundaries are polygons for now
-        let polygon: Polygon = boundary.try_into()?;
-        let multi_polygon: MultiPolygon = polygon.into();
+        let input_bytes = std::fs::read(&self.pbf_path())
+            .context(format!("unable to read '{}'", self.pbf_path()))?;
```
michaelkirk, line 102: I've long known about `context` in theory, but I finally used it today! Wow!

Expect more of this in the future! 

In case you aren't familiar... I find rusts `?` easy to write, but it can make tracking down where the error originated tricky. Stack traces are great for this, but only happen when you have a crash, not a `?` propogation.

Context will add this text to the error that gets propagated, so when you fail or crash at an outer level, you'll still see this context string.

michaelkirk: Of course now that I think about it, this particular code is all test code, so maybe we should just be `unwrap`-ing everywhere anyway... but I think you get the idea.

dabreegster: I also haven't used it too much, but agree it's super helpful! For the particular case of getting filenames into errors, I've also used https://docs.rs/fs-err/latest/fs_err/index.html before

dabreegster (approved):
Nice speedup!

backend/src/boundary_stats.rs
dabreegster: Thanks for the explanation!

Issue #156, opened 2025-02-20
prioritization metric: stats 19


michaelkirk:
This is a point data set, which we should be able to reasonably easily count per area.

Incident happening *on* the neighbourhood boundary present an interesting challenge. 

I understand that we want to include incidents on the perimeter in the neighbourhood's count. Due to numerical precision it's arbitrary whether a point approximately "on" the line will be inside or outside the shape, so we might need to slightly buffer the geometric shapes to account for this. 

(Perhaps a good reason to take a look at the new i_overlay buffer integration!)

On the call there was support for a simple calculation of incident density, but not distinguishing between other attributes of the incidents (like severity).

dabreegster:
I'll start this as a contextual layer soon, to get the data. It requires a little bit of processing, but it's already done for an ATE project.

michaelkirk:
With #171 and #174, I believe the only thing remaining for stats19 is a CNT-approved methodology (or approval of what we're already doing).

michaelkirk:
The only thing left to do here has to do with labeling:

1. rather than 1-100 (percentile) bucket to quintile (1-5)
2. Label "most deprived / least deprived"

michaelkirk:
Whoops, ignore my previous comment, wrong metric!

Remaining work to do for stats19: update scale to be dynamic equal interval rather than our custom "logarithmic buckets". So the scale will be different depending on which LAD you're looking at.

One thing that's not obvious to me - what's the unit of analysis that we will use for this bucketing? My guess: the original autogenerated boundaries?

dabreegster:
Good question... I think the original autogen boundaries make sense. Nothing else really does. It does feel a bit odd to take a dependency on how we measure these, though...

PR #171, opened 2025-02-23
Starts a stats19 prioritization metric

dabreegster:
#156

1) Filter out collisions not involving a ped/cyclist upfront
2) Generalize the "populate zone" file to a "context data" file, and include individual stats19 points in it
3) Define a simple metric counting the number of those points completely contained by a boundary
4) Refactor a bit of the frontend code to make adding more metrics easier

This is just starting a proof-of-concept. We'll want to:

- buffer the boundaries and find collisions on the perimeter road
- define the metric in terms of density, not count?
- adjust the color scale / limits completely


backend/benches/boundary_stats.rs
dabreegster, line 1: I haven't rerun this benchmark to see how slow point matching is. I think I'm missing `tests/dundee.geojson` -- did you mean to commit it?

backend/src/test_fixtures.rs
```diff
@@ -74,8 +74,9 @@ impl NeighbourhoodFixture {
         }
     }
 
-    pub fn population_zones_path(&self) -> Option<String> {
+    pub fn context_data_path(&self) -> Option<String> {
         if self.is_cnt {
+            // TODO Rename this
```
dabreegster, line 79: Before I change the scripts and cloudflare, do you like "context data" or any better ideas?

data_prep/scotland/get_reference_layers.sh
```diff
@@ -92,7 +92,7 @@ function stats19 {
   download_to_subdir input https://data.dft.gov.uk/road-accidents-safety-data/dft-road-casualty-statistics-casualty-1979-latest-published-year.csv
   download_to_subdir input https://data.dft.gov.uk/road-accidents-safety-data/dft-road-casualty-statistics-collision-1979-latest-published-year.csv
 
-  cargo run --release
+  cargo run --release -- --only_pedestrians_and_cyclists
```
dabreegster, line 95: I made a quick change in the other repo to do this

web/src/AutoBoundariesMode.svelte
```diff
@@ -79,6 +88,46 @@
     }
     return x;
   }
+
+  function fillColor(
```
dabreegster, line 92: I thought about sharing more code with pick neighbourhood, but:

1) We have different coloring for "none" between the two modes
2) The popup behavior is different

Duplicating doesn't seem too bad

michaelkirk (approved):
The code all looks great!

I know you said this is only a POC, so I don't think this is news to you, but the current output is very flat:

Screenshot 2025-02-24 at 12 02 31

Screenshot 2025-02-24 at 11 57 31


I'm sure we'll revisit based on feedback anyway, but I think using counts/area (collision density) would likely be "better behaved" as a metric.
And tweaking the limits around so they at least look plausible in one LAD would be helpful for demoing this feature. I can take a quick pass at that after merging if you'd like.

data_prep/scotland/src/bin/generate_prioritization.rs
```diff
@@ -94,3 +103,19 @@ impl PopulationZoneInput {
         Ok(iter.collect())
     }
 }
+
+// Ignore all properties
+#[derive(Deserialize)]
+struct Stats19Input {
+    #[serde(deserialize_with = "geojson::de::deserialize_geometry")]
+    geometry: Point,
```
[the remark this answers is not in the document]
dabreegster: Correct

backend/src/create.rs
```diff
@@ -154,7 +155,12 @@ pub fn create_from_osm(
                     population_zone,
                 }
             })
-            .collect()
+            .collect(),
+        stats19_collisions: context_data_wgs84
+            .stats19_collisions
+            .into_iter()
+            .map(|pt| graph.mercator.to_mercator(&pt))
```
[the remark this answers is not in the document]
dabreegster: That's exactly right. In abstreet, the purpose of `LngLat` vs `Pt2D` was to distinguish. The lack of type safety around this hasn't been a real problem for me so far, though -- convention and only handling WGS84 "on the edge" has been OK.

dabreegster:
Thanks for the review! I switched to a density metric and made Edinburgh look interesting:
image
More than happy to iterate on this later. Maybe a log scale would make more sense, for instance

michaelkirk:
> I switched to a density metric and made Edinburgh look interesting

My own spot checks look much more plausibly useful after this change. Thanks for doing that!

PR #172, opened 2025-02-25
Iterate on the context layers UI

### LINKED ISSUE #109: cleanup contextual layers UI
Tracking the work left by #107 

> The controls are just about as painfully awkward as I could make them:
[Screencast from 07-02-25 16:38:28.webm](https://github.com/user-attachments/assets/5c6ad487-cd2c-4e24-b2b6-9d52103b7549)
MapLibre control styling is really fighting pico here! My intention here is to get lots of the work started quickly, and figure out how to polish the UI as we go along. There's some UX work happening in the parallel NPW project, and when it's ready in a few weeks, the same ideas could be applied here. In the meantime, please do suggest any ideas for improvements.

We'll either have to figure out how to get pico to play nice, or build something non-pico.

michaelkirk:
FIXES #109

A lot of this is subjective, so feel free to raise issues you see.

Overall, I think this is a more compact, while still usable and uniform way to display the contextual layers.

**before:**

https://github.com/user-attachments/assets/d14590ba-8c8e-4f39-b685-016c651ed9c1

**after:**

https://github.com/user-attachments/assets/0dec4a9a-1085-444d-9e3b-675e58814360


web/src/common/ContextLayerButton.svelte
```diff
@@ -0,0 +1,41 @@
...
+
+<button
+  style="display: flex; align-items: center; justify-content: leading; gap: 8px; text-align: left;"
+  class="secondary"
+  on:click={() => (show = !show)}
+>
+  <input type="checkbox" bind:checked={show} />
+  {label}
+  {#if $$slots.help}
+    <HelpButton color="var(--pico-secondary-inverse)">
```
dabreegster, line 16: The help button nested inside the layer button means that when you click help, it toggles the layer on and off. This feels unintuitive to me, but happy to go with this for now and discuss/iterate later. Maybe just ending the outer `<button>` after `{label}` would be enough, I'll try later.

web/src/context/ContextualLayers.svelte
```diff
@@ -14,17 +14,29 @@
...
+        {/if}
+      </div>
     </button>
 
     <div
-      style="display: flex; flex-direction: column"
-      style:visibility={expand ? "visible" : "collapse"}
+      style="flex-direction: column"
+      style:display={expand ? "flex" : "none"}
+      style:background-color="#515f7A"
```
dabreegster, line 39: It's a bit hard to see some of the legend text against the grey; maybe we could try white foreground text?
image

dabreegster (approved):
Thank you, this is all a big improvement!

One issue I haven't figured out yet is making the `<Modal>` used for the help to take/center in the entire window, not just the right bar. This is probably going to be a change to that component in svelte-utils (or a temporary copy of it over here)

michaelkirk:
Thanks for the feedback! I think I've addressed it all in #173

PR #173, opened 2025-02-25
More layer fixups

michaelkirk:
Followup to #172:

Draft because we should merge https://github.com/a-b-street/svelte-utils/pull/3 first.

But you can better see how https://github.com/a-b-street/svelte-utils/pull/3 is used in this PR.

## increased legend background contrast for legibility

**before:** 

Screenshot 2025-02-25 at 12 18 17

**after:**

Screenshot 2025-02-25 at 12 18 32

## layer help modal is now presented modally, not squashed into the layer container
**before modal presentation:**

Screenshot 2025-02-25 at 12 11 33

**after modal presentation:**

Screenshot 2025-02-25 at 12 10 57

## clicking layer "help" to present modal no longer triggers layer visibility 

Clicking to dismiss the modal *also* toggled the visibility, so it kind of cancelled itself out enough for me to not notice!

I recommend you review with whitespace diff toggled off since there's a lot of indentation changes.


web/src/About.svelte
```diff
@@ -3,49 +3,47 @@
...
-        OpenStreetMap
-      </a>
-      data.
-    </p>
-    <center>
-      <button on:click={() => notNull(dialog).close()}>Start!</button>
-    </center>
-  </Modal>
-{/if}
+<Modal bind:show={$showAbout} let:dialog>
```
dabreegster, line 6: nit: I think `let:dialog` can go away now?

dabreegster (approved):
Thank you! Huge improvement. LGTM after merging the other PR

michaelkirk:
squash merged in 34d632c

web/src/About.svelte
michaelkirk: Good catch - done!

PR #174, opened 2025-02-26
buffer neighbourhood stats

michaelkirk:
It's about 10-15% slower to include buffering.  But it does significantly affect the results. Do you think it's worth it?

Personally I think we should. I think there is some performance work I can (would like to) do on the geo side to speed this up a bit, plus we can always pull out the buffering in the worst case if it proves intolerably slow.

**before:**

Screenshot 2025-02-25 at 16 06 29

**after:**

Screenshot 2025-02-25 at 16 08 29

## autogenerated screen

Sorry, it's hard to tell, but I'm hovering the same boundary.

**before:**

Screenshot 2025-02-25 at 16 07 03

**after:**

Screenshot 2025-02-25 at 16 10 20

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -82,7 +82,6 @@ dependencies = [
  "fast_paths",
  "geo",
  "geojson",
- "i_float",
  "i_overlay",
  "itertools 0.14.0",
  "log",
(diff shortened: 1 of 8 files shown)


michaelkirk:
I have some experiments that would be good to keep in the PR for posterity, but please squash if you merge into main.

dabreegster (approved):
The time increase isn't that bad at all, and this is definitely necessary, so let's go for it! Thanks for initiating this exciting first step into the world of geo buffering!

PR #175, opened 2025-02-26
Visualize when perimeter roads are included by buffering boundary.

michaelkirk:
This is a little UI affordance inspired by my desire to try out more of the geo buffering integration, but I think it's helpful!

---
**before:** The boundary falls directly *on* the perimeter centerline (by definition). When looking at the editor it's ambiguous if the perimeter is included or not.

Screenshot 2025-02-26 at 12 19 44

---

**after:** Let's make that explicit, depending on the value of `edit_perimeter_raods`, by growing or shrinking the buffer around the neighbourhood.

**exclude perimeter:**
Screenshot 2025-02-26 at 12 18 47

**include perimeter:**
Screenshot 2025-02-26 at 12 18 52


dabreegster (approved):
This is incredibly helpful/nice! We can get feedback tomorrow, but I think maybe even buffering outwards to show the perimeter all of the time could be useful, whether or not those roads are editable. It emphasizes the area better.

michaelkirk:
Whoops. Test fixes forthcoming.

michaelkirk:
>  It emphasizes the area better.

It does look nice. Maybe "edit perimeter roads" should be on by default?

dabreegster:
> Maybe "edit perimeter roads" should be on by default?

It's maybe worth asking feedback about this again. I think:

- being able to filter/edit perimeter roads is _sometimes_ useful. Maybe it should be on by default.
- buffering to show the perimeter is probably always nice
- but the shortcuts calculation when including perimeter roads is notably slower and the results still feel subtly weird / not emphasizing the same design process

michaelkirk:
Thanks for breaking down some of the issues. I have some thoughts! I'm going to file a separate issue for discussion shortly.

PR #178, opened 2025-02-28
Disable 3D buildings

dabreegster:
Before, at high zoom, 3D building extrusions are a little visually noisy. I can't find a bad case right now, but sometimes the shadow even obscures roads that we're trying to color.
image

After, just use the flat 2D all the time:
image

### DIFF
--- a/web/src/App.svelte
+++ b/web/src/App.svelte
@@ -6,7 +6,7 @@
   import ContextualLayers from "./context/ContextualLayers.svelte";
   import "@picocss/pico/css/pico.conditional.jade.min.css";
   import initLtn from "backend";
-  import type { Map } from "maplibre-gl";
+  import type { Map, StyleSpecification } from "maplibre-gl";
   import { init as initRouteSnapper } from "route-snapper-ts";
   import { onMount } from "svelte";
   import {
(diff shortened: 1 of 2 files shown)


web/src/App.svelte
```diff
@@ -91,6 +91,25 @@
...
   }
+
+  async function getStyle(
+    basemap: string,
+  ): Promise<StyleSpecification | string> {
+    if (basemap != "streets-v2") {
+      return `https://api.maptiler.com/maps/${basemap}/style.json?key=${maptilerApiKey}`;
+    }
+
+    // streets-v2 uses a fill-extrusion layer for 3D buildings that's very distracting. Remove it, and make the regular buildings layer display at high zoom instead.
```
dabreegster, line 102: I thought about downloading this JSON, modifying it manually, and storing in git, but for these two simple changes, that feels wrong. The basemap styles don't have breaking changes without a version bump.

dabreegster:
image
Aha, here's an example of 3D buildings overlapping roads. A bit zoomed out.

michaelkirk (approved):
I can see how the 3-d buildings will sometimes interfere with all the other visualization we're trying to do on the map, so I agree with getting rid of it.

PR #179, opened 2025-02-28
When two roads meet, draw the one with more shortcuts on top

dabreegster:
Road center-lines meet at the same point, so when the lines are thickened in maplibre, the polygons overlap a bit.

Before, the lines are drawn in some arbitrary order (based on road IDs, just the order we scrape OSM data). White deadends with no shortcuts partly cover up red roads:
image

After, it looks a little cleaner in some places:
image


michaelkirk (approved):
Nice and simple!

Issue #176, opened 2025-02-26
Get rid of "Include Perimeter roads" toggle, always edit perimeter roads.


michaelkirk:
In #175 [I proposed](https://github.com/a-b-street/ltn/pull/175#issuecomment-2686126815):

> Maybe "edit perimeter roads" should be on by default?

@dabreegster responded:
> It's maybe worth asking feedback about this again. I think:
> - being able to filter/edit perimeter roads is sometimes useful. Maybe it should be on by default.
> - buffering to show the perimeter is probably always nice
> - but the shortcuts calculation when including perimeter roads is notably slower and the results still feel subtly weird / not emphasizing the same design process

I've got some thoughts about this, but it didn't feel like it necessarily needed to block #175 from merging. I wanted to make this issue to hash out the details.

> buffering to show the perimeter is probably always nice

I agree that it *looks* nice. 

However, the primary motivation for #175  was about usability. I'm trying to avoid complaints like:

"Why can't I edit this piece of the road network even though it looks exactly like all the other roads on the screen?"
"You need to check the box in the panel first."

Buffering like this allows us to have the mask be a visual indicator of "what is editable" which is otherwise missing (except on hover). So unless we come up with some other design treatment that solves this problem, I'd prefer to keep the perimeter "in the dark" when it's not editable.

> but the shortcuts calculation when including perimeter roads is notably slower and the results still feel subtly weird / not emphasizing the same design process

Regardless of if perimeter editing is allowed, if we're ever counting trips *along* the perimeter as shortcuts, I think we should change that. 
By definition, to be a shortcut you must be cutting "into" the interior roads - right? It seems very intuitive to me, but maybe it's more complicated than that - am I missing something? 

If you can point me to any examples/screen recordings that highlight the current problems, that would be helpful.

Maybe the desired behavior is something we can clarify on the call tomorrow.

> being able to filter/edit perimeter roads is sometimes useful. Maybe it should be on by default.

Just for context, my own guess is that the most likely treatments applied to the perimeter road will be adding turn restrictions, and less so editing road directions and adding modal filters. Does that seem right?

If we end up not counting "shortcuts" on the perimeter as I'm suggesting, then I'm left wondering: Why do we want to prevent users from editing the perimeter until they click a checkbox? Currently for me it's a bit inexplicable.

In summary, if we:
- change the shortcut logic to only include interior trips
- always allow editing perimeter roads (thus remove the checkbox)

[...]

PR #180, opened 2025-02-28
Edit turn restrictions

dabreegster:
A big step on #30! I'll comment on parts of the code with some TODOs and move them into the issue.


backend/src/map_model.rs
```diff
@@ -46,6 +46,11 @@ pub struct MapModel {
     pub modal_filters: BTreeMap<RoadID, ModalFilter>,
     pub diagonal_filters: BTreeMap<IntersectionID, DiagonalFilter>,
 
+    /// Indexed by IntersectionID. For each intersection, a list of (from, to) roads that are not
```
dabreegster, line 49: Mirroring the structures above for editable things, rather than bundling with each intersection. This is expected to be a pretty sparse list -- most intersections don't have TRs.

backend/src/map_model.rs
```diff
@@ -1024,6 +1141,7 @@ pub enum Command {
     SetModalFilter(RoadID, Option<ModalFilter>),
     SetDiagonalFilter(IntersectionID, Option<DiagonalFilter>),
     SetTravelFlow(RoadID, TravelFlow),
+    SetTurnRestrictions(IntersectionID, Vec<(RoadID, RoadID)>),
```
dabreegster, line 1144: Every time I design savefile formats or think through how to implement the command pattern for undo, I hit this question about objects with lists. We could also have a command for "add" and "delete" one individual TR. In our case, the list order here doesn't matter.

I started with this because it was simpler and the choice doesn't seem to matter here, but open to opinions

backend/src/movements.rs
```diff
@@ -78,44 +76,144 @@ impl MapModel {
...
+                if let Some(name) = to.tags.get("name") {
+                    f.set_property("name", name.clone());
+                }
+                features.push(f);
+            }
+        }
+        GeoJson::from(features)
+    }
+
+    /// Find (intersection, from, to) to represent a turn restriction
```
dabreegster, line 126: How do we robustly represent turn restrictions and make sure reasonable OSM edits don't break things? Simplest approach seems to be

1) Store a point for the intersection, and hope the closest one doesn't change much.   (This might be wrong if somebody does something like turn a single node with dual carriageways intersecting into 4 nodes, but we're kind of hopeless in that situation anyway)
2) Store the bearing of the road pointing into the intersection and then away from the intersection, calculate that bearing again, find the closest match by rotation

dabreegster: This deserves some careful unit tests at some point, along with the other logic for loading savefiles

backend/src/movements.rs
```diff
@@ -157,6 +255,13 @@ fn classify_relative_bearing(abs_bearing1: f64, abs_bearing2: f64) -> &'static s
     }
 }
 
+// Given two angles in [0, 360), return the smallest rotation between them
```
dabreegster, line 258: I am so bad at reasoning about angles, but I think this is simple and correct, and the unit tests agree...

backend/src/movements.rs
dabreegster: It is a little weird, maybe, to use GJ as our output format and have multiple points in exactly the same place for representing multiple turn restrictions. We could also have one point with an array, but then what if you delete an existing TR somewhere and add a new one at the same intersection? It'd be inconsistent with our savefile format to express the "new thing" vs "deleted existing" in a different way...

web/src/edit/NeighbourhoodMode.svelte
```diff
@@ -88,26 +118,70 @@
...
+    } else if (action.kind == "oneway") {
       $backend!.toggleTravelFlow(f.properties!.road);
       $mutationCounter++;
+    } else if (action.kind == "turn_restriction") {
+      action.from_road_id = f.properties!.road;
+      action.from_road_name = f.properties!.name || "unnamed road";
+      action.possible_targets = $backend!.getTurnRestrictionTargets(
+        f.properties!.road,
+      );
+      // TODO The blue hover state gets stuck
```
dabreegster, line 133: Click a road to start making a TR, then click the map to deselect. Blue hover stays for reasons unknown. If you hover over any road, then it resets. I'll fix later, unless you spot the problem easily in review

web/src/edit/NeighbourhoodMode.svelte
```diff
@@ -88,26 +118,70 @@
...
   }
 
+  function deleteTurnRestriction(e: CustomEvent<LayerClickInfo>) {
+    let f = e.detail.features[0];
+    $backend!.deleteTurnRestriction(
+      f.properties!.intersection,
+      f.properties!.from_road,
+      f.properties!.to_road,
+    );
+    $mutationCounter++;
```
dabreegster, line 177: A bug with the preview lines...

1) Hover on a TR icon. You see the red lines.
2) Click to delete
3) Those lines remain until you move the mouse

I tried some stuff to force recalculating that GJ when mutationCounter changes, but no luck yet

web/src/edit/NeighbourhoodMode.svelte
```diff
@@ -248,36 +325,54 @@
       </mark>
     {/if}
 
-    <div style="display: flex; justify-content: space-between;">
+    <div style="display: flex; justify-content: space-between">
```
dabreegster, line 328: I switched to icon buttons to make things fit, as discussed.

- bug 1: the tooltips get clipped by the map canvas. Not sure how to fix pico zordering without modifying its stylesheets, but probably a `:global(some class)` should do the trick
- bug 2: Zoom your browser in / out. The icons always stretch and get absurdly big or small. The responsive behavior is definitely kind of wrong. Getting the icon images to have the same size would probably be a good start.

web/src/edit/NeighbourhoodMode.svelte
```diff
@@ -372,12 +483,48 @@
...
+        <Popup openOn="hover">Click to delete</Popup>
+      </div>
     </ModalFilterLayer>
 
-    {#if action == "freehand-filters"}
+    {#if action.kind == "freehand-filters"}
       <FreehandLine map={notNull($map)} on:done={gotFreehandLine} />
     {/if}
+
+    {#if action.kind == "turn_restriction"}
```
dabreegster, line 502: image
The icon in the popup is indeed helpful, thanks for that idea!

dabreegster (commented):
Two more observations:

- It's a little bit weird, but you can create TRs where the second road is totally outside the drawn boundary, because of how the targets work.
- Much of the time, the interesting TRs will be exiting/entering the neighbourhood using the perimeter road. So that's more motivation for #176 -- by default, TRs should be able to start from a perimeter road without ticking a box.

backend/src/map_model.rs
```diff
@@ -423,6 +471,11 @@ impl MapModel {
                 self.travel_flows.insert(r, dir);
                 Command::SetTravelFlow(r, prev)
             }
+            Command::SetTurnRestrictions(i, restrictions) => {
+                let prev = std::mem::take(&mut self.turn_restrictions[i.0]);
```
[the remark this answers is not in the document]
dabreegster: Done

backend/src/movements.rs
```diff
@@ -78,44 +76,144 @@ impl MapModel {
...
+                road2.pt_near_intersection(0, self.id).into(),
+            ),
+        )
     }
+}
+
+impl Road {
+    fn pt_near_intersection(&self, offset: usize, i: IntersectionID) -> Point {
+        // If the road is long enough, offset from the intersection this much
+        let distance_away = 5.0 * (1 + offset) as f64;
```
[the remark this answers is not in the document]
dabreegster: Great idea, this is much more clear

web/src/edit/NeighbourhoodMode.svelte
dabreegster: Hmm, should we do the same thing in the popup for modal filter icons then? Added as something to revisit in #30

dabreegster (commented):
> I noticed after placing a turn restriction, I'm switched back to the "place a modal" tool.

Fixed. The only case where we switch back to filter is after drawing a freehand line. That one seems appropriate to me; the freehand tool is meant for quickly fixing shortcuts in one direction by just drawing a huge line. Usually the action after that is to adjust the filters created, find redundant ones, etc.

PR #184, opened 2025-03-04
Touch up tool palette UI

michaelkirk:
A small touchup to clarify the "selected" state of the active tool.

**before: selected item is de-emphasized due to it being "disabled"**

https://github.com/user-attachments/assets/07392d54-ed9b-4c8c-bb5f-1aa573b354da

**after:**

https://github.com/user-attachments/assets/950d9270-2e35-4671-837a-3e51817a7085

### DIFF
--- a/web/src/edit/NeighbourhoodMode.svelte
+++ b/web/src/edit/NeighbourhoodMode.svelte
@@ -325,49 +325,50 @@
       </mark>
     {/if}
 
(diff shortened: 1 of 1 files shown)


dabreegster (approved):
Thank you, this looks great! And I'm learning some CSS techniques I didn't know about... `gap` and `object-fit` are quite useful

PR #185, opened 2025-03-05
update geo

michaelkirk:
Note this includes the as-of-yet unmerged https://github.com/georust/geo/pull/1321 and https://github.com/georust/geo/pull/1322

Primarily this gets us a nice perf boost in the stats calculation:
  
    build stats: Hilltown in LAD_Dundee City
                            time:   [890.95 µs 894.88 µs 899.45 µs]
                            change: [-34.332% -34.080% -33.830%] (p = 0.00 < 0.05)
                            Performance has improved.
    
    generate auto boundaries (and stats) for all of LAD_Dundee City
                            time:   [192.06 ms 192.32 ms 192.67 ms]
                            change: [-61.269% -61.036% -60.823%] (p = 0.00 < 0.05)
                            Performance has improved.
    
And some modest wins elsewhere...
    
    build map_model: bristol
                            time:   [83.446 ms 83.560 ms 83.688 ms]
                            change: [-3.4679% -2.8247% -2.2423%] (p = 0.00 < 0.05)
                            Performance has improved.

### DIFF
--- a/backend/src/create.rs
+++ b/backend/src/create.rs
@@ -1,7 +1,8 @@
 use std::collections::{BTreeMap, BTreeSet, HashMap};
 
 use anyhow::Result;
-use geo::{Coord, LineInterpolatePoint, LineString, MultiPolygon, PreparedGeometry};
+use geo::line_measures::InterpolatableLine;
+use geo::{Coord, Euclidean, LineString, MultiPolygon, PreparedGeometry};
 use osm_reader::{NodeID, OsmID, RelationID, WayID};
 use petgraph::graphmap::UnGraphMap;
 use rstar::{primitives::GeomWithData, RTree};
@@ -360,7 +361,7 @@ fn apply_existing_filters(
             let pt = map
                 .get_r(r)
                 .linestring
-                .line_interpolate_point(percent)
+                .point_at_ratio_from_start(&Euclidean, percent)
                 .unwrap();
             map.add_modal_filter(pt.into(), Some(vec![r]), filter);
         }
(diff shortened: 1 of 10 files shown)


backend/src/create.rs
```diff
@@ -360,7 +361,7 @@ fn apply_existing_filters(
             let pt = map
                 .get_r(r)
                 .linestring
-                .line_interpolate_point(percent)
+                .point_at_ratio_from_start(&Euclidean, percent)
```
michaelkirk, line 364: This is the new api from https://github.com/georust/geo/pull/1321 - it's more explicit about metric space, and also has some new APIs which you can see further down.

backend/src/movements.rs
```diff
@@ -211,20 +212,15 @@ impl Road {
...
-                1.0 - (meters_away / len)
-            };
-            return self.linestring.line_interpolate_point(pct).unwrap();
+        if self.src_i == i {
+            self.linestring
+                .point_at_distance_from_start(&Euclidean, meters_away)
+                .unwrap()
+        } else {
+            self.linestring
+                .point_at_distance_from_end(&Euclidean, meters_away)
```
michaelkirk, line 221: Here's a good use of the new api - we're using `distance` not `ratio` and computing from the `end` not the `start`.

Note that this method already clamps the output to the linestring.

dabreegster (approved):
More awesome perf boosts, and the new interpolation APIs are so much more clear! I'll leave some comments on the upstream PR soon. Distinguishing clearly between distances and ratios is very nice; it might work well also for the line split trait API.

PR #186, opened 2025-03-05
Adjust SIMD legend into quintiles and tweak copy

### LINKED ISSUE #136: Show SIMD per neighbourhoood in the neighbourhood picker

michaelkirk:
FIXES #136 (mostly done before, but these are the final touches)

Some small tweaks based on our last meeting.

**before:**
simd-before

**after:**
simd-after

### DIFF
--- a/web/src/common/SequentialLegendBucketed.svelte
+++ b/web/src/common/SequentialLegendBucketed.svelte
@@ -1,0 +1,22 @@
+<script>
+  export let colorScale;
+  export let buckets;
+  export let decimalPlaces = 0;
+</script>
+
+<div style="display: flex">
+  {#each colorScale as color}
+    <span style="background: {color}; width: 100%; border: 1px solid black;">
+      &nbsp;
+    </span>
+  {/each}
+</div>
+<div style="display: flex; justify-content: space-around;">
+  {#each buckets as bucket}
+    {#if decimalPlaces > 0}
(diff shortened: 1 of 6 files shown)


web/src/common/SequentialLegendBucketed.svelte
```diff
@@ -0,0 +1,22 @@
+<script>
```
michaelkirk, line 1: This is a fork of SequentialLegend from svelte-utils, but the labels fall **on** the buckets rather than **between** the buckets (consequently, there's one less label).

dabreegster: Nice! I might upstream this and use in NPW too.

Missing TS types, but I'll add them after merging the PR stacks to avoid any rebase headache.

web/src/context/Population.svelte
```diff
@@ -78,8 +85,8 @@
       <p>
         Data zone {props.id}
         has {props.population.toLocaleString()}
-        people, and a SIMD rank of {props.imd_rank}, putting it in the {props.imd_percentile}
-        percentile.
+        people, and a SIMD rank of {props.imd_rank}, making it less deprived
+        than {props.imd_percentile}% of data zones.
```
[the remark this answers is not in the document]
dabreegster: This reads fine to me. We could ask if "more deprived" would be preferable/easier to read.

dabreegster (approved):
Awesome, thanks!

Issue #181, opened 2025-03-01
modal filter problems in early and later LTN tools Harborne East LTN


Carantouhil:
I have been using the earlier and the latest versions to design a LTN, but mostly the earlier version. Periodically impact data is not consistent through repeat runs. This is an example where there errors with 5 modal filters using the original downloaded LTN tool. 

![Image](https://github.com/user-attachments/assets/14f16640-936c-4451-8fdd-8220fffad2ce)

I have repeated the design with the latest web-based LTN tool. I then then checked Routing and am seeing similar bugs where route pass through filters. 

![Image](https://github.com/user-attachments/assets/27d066c8-fc72-4be0-8490-ff4c9098f4c2)

Animate shortcuts shows an error with one filter on Fitzroy Avenue: 

https://github.com/user-attachments/assets/c707648b-213a-460a-b1c5-3972157c9019

Any help would be greatly appreciated

I have geoson files, but can't add here

Carantouhil:
Thanks
[ltn_Harborne East Vio.txt](https://github.com/user-attachments/files/19039179/ltn_Harborne.East.Vio.txt)

dabreegster:
![Image](https://github.com/user-attachments/assets/e682ae48-1d98-4c02-a97e-534144ca474c)
To visualize the 2nd problem better, when I move the A marker so that it snaps to that little deadend road, then things work as expected

Carantouhil:
thanks - that's great!

dabreegster:
#266 also closed this. Both of these problems should be fixed now

PR #189, opened 2025-03-06
Fix routes incorrectly ending on a road with a modal filter

dabreegster:
Cici spotted quite a notable bug:
image
An interior road right on the edge with a filter on it should not have any shortcuts! View shortcut mode reveals the problem is routes ending there:
image

Previously, I think we got away with this because of https://github.com/a-b-street/ltn/blob/00fb50bbedf1cb37606a54de99a91398a85e4c3e/backend/src/route.rs#L63. Roads with filters weren't part of the CH graph at all. But the change to `(road, direction)` pairs and handling all roads connected to the start/end intersection somehow let this slip in.

### DIFF
--- a/backend/src/map_model.rs
+++ b/backend/src/map_model.rs
@@ -158,6 +158,9 @@ impl Intersection {
             if to_road.id == from_road.id {
                 return None;
             }
+            if router_input.has_modal_filter(to_road.id) {
+                return None;
+            }
             if router_input
                 .turn_restrictions(self.id)
                 .contains(&(from_r, to_road.id))
--- a/backend/src/route.rs
+++ b/backend/src/route.rs
@@ -210,7 +210,7 @@ impl Route {
 #[cfg(test)]
 mod tests {
     use super::*;
-    use crate::osm_tests::load_osm_xml;
+    use crate::{osm_tests::load_osm_xml, FilterKind};
 
     fn r(road_id: usize) -> RoadID {
         RoadID(road_id)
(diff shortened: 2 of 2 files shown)


backend/src/map_model.rs
```diff
@@ -158,6 +158,9 @@ impl Intersection {
             if to_road.id == from_road.id {
                 return None;
             }
+            if router_input.has_modal_filter(to_road.id) {
+                return None;
```
dabreegster, line 162: I think we will relax this later to solve part of #181. When a route starts or ends on a filtered road, we can travel only one direction along it, depending on the start position.

backend/src/route.rs
```diff
@@ -233,12 +233,24 @@ mod tests {
...
-        let map = load_osm_xml("simple_four_way_intersection");
+        let mut map = load_osm_xml("simple_four_way_intersection");
         let Route { steps } = map.router_before.route_from_roads(r(3), r(2)).unwrap();
         assert_eq!(
             steps,
             vec![(r(3), Direction::Backwards), (r(2), Direction::Backwards)]
         );
+
+        // Filter r2
+        map.add_modal_filter(
```
michaelkirk, line 244: Thank you for adding a test! I think routing could be a bit of a whack-a-mole problem, so it's good to make sure we're not regressing.

michaelkirk (approved):
LGTM! 

Yeah, this was just a bad bug in my implementation.

backend/src/route.rs
dabreegster: Easy to add tests once there's at least one :)  Let the virtuous cycle continue!

PR #190, opened 2025-03-06
Remove SIMD popup

michaelkirk:
Some of the items that came out of our meeting were very small and straight forward. 

For those I'm going to open PRs directly rather than first write up an issues. 

LMK if you'd like me to do otherwise - I should be done with them within the hour.


michaelkirk:
Sorry, I've just now pushed some related commits, addressing related trivial issues.

dabreegster (approved):

PR #204, opened 2025-03-06
Remove holes (lakes) from generated boundaries

### LINKED ISSUE #145: Confusing artifacts in autogenerated boundaries

michaelkirk:
FIXES #145 (just the final bit)

I hadn't really considered this, but this actually has some (small) correctness implications for our prioritization aggregation:

**before:**
Screenshot 2025-03-06 at 13 10 48

**after:**
Screenshot 2025-03-06 at 13 13 50


dabreegster (approved):
Nice and simple! Results look much better

Issue #104, opened 2025-02-07
Specialize for CNT


dabreegster:
The general LTN tool is being specialized in a few ways for a use case in Scotland, called Connected Neighbourhoods Tool (CNT). This issue tracks technical debt and concerns about keeping one app serving the original use case of "import anywhere from OSM" and the special Scottish areas.

- [ ] When the user starts in CNT, then later does "Choose project", should we redirect to the landing page?
- [ ] Should all OSM areas be namespaced, so the URL logic can be less custom?
- [ ] Should all savefiles be namespaced as `ltn_tool/boundary_name/filename`?

michaelkirk:
My understanding of the underlying issue is that we have two different landing pages that solve the same task of viewing/managing your projects.

1. ltn/cnt.html is a landing page specialized for CNT projects.
2. ltn/index.html, is a general purpose landing page for users outside of Scotland.


Speaking *only* from a UX perspective, and not thinking at all about the ramifications on the implementation, CNT users should be focused on CNT projects. 

We shouldn't make it possible (or at least, not make it so easy) for CNT users to accidentally wind up in the "non-CNT" (general purpose) LTN tool.

Does that seem like a desirable UX, or am I missing an important use case?

dabreegster:
That's spot on, CNT users winding up in the general-purpose LTN tool or vice versa does not make sense. Arguably and maybe ultimately there should be two totally separate deployments, especially if the branding/styling changes. In the short term...

- We could make the URL structure less confusing maybe? Going from /cnt.html to /index.html?project=ltn_cnt/something isn't intuitive
- "Choose project" in CNT could go back to cnt.html?

michaelkirk:
I got somewhat of a wild hair and started working on this today - I had a "moment of clarity" after demoing the project to someone, and came up with the basic idea.

You can see the WIP here: https://github.com/michaelkirk/ltn/pull/new/mkirk/reconcile-landing

The overall idea is really simple: add a toggle to switch between "Global" and "CNT", and get rid of the cnt.html page.

https://github.com/user-attachments/assets/b2abc882-f5e1-4a58-bc79-63dbdd1cf4b2

As it's still WIP, I haven't figured out how much further I'll have to "chase" this change throughout the code base.

But what do you think of the basic flow?

dabreegster:
[...]

Issue #195, opened 2025-03-06
Error loading saved project


michaelkirk:
<img width="505" alt="Image" src="https://github.com/user-attachments/assets/1f3dc6e2-399f-4eeb-9c75-b4c75c3e11c9" />

This happened while trying to open Michael N's example neighbourhood. I haven't had a chance to investigate it yet.

dabreegster:
We detect CNT vs other built-in areas from the local storage key's prefix. The logic to load a saved project doesn't set the correct prefix for CNT. We need to detect it and do that.

Relatedly, on the CNT landing page, we need this same "load saved project" feature. I'm not sure what should happen if the user loads a project on the "wrong" landing page, but I think the file contents should win. In other words, if you load  a CNT project on index.html, treat it as CNT. If you load a regular project on cnt.html, treat it as normal / not CNT.

dabreegster:
Almost done...

- [ ] Rename cnt.html to cnt/index.html
- [ ] When switching appFocus, immediately rewrite the URL in history

Issue #169, opened 2025-02-22
Couldn't open project: Unknown kind in savefile: direction


XioNoX:
I see that lots of changes happened, that's awesome !

However I can't open my saved projects. When I click on them I get a popup that says `Couldn't open project: Unknown kind in savefile: direction`.
Hopefully it's an easy one to fix, otherwise I can recreate the various projects.

Thanks,

dabreegster:
Yep, #119 made a change. For the next month or so, there'll be more heavy development and breaking changes. #138 tracks us establishing a proper v1 of the save file format and making sure to preserve backwards compatibility after that point.

In the meantime, the easy fix is to replace `direction` with `travel_flow` in the savefile. Actually doing this is a little clunky:

1) Ctrl+Shift+I to open browser devconsole
2) Depending on your browser, find the Memory > Local Storage section
3) Find the key corresponding to this file (or do it for all of them)
4) Copy the value to a text editor
5) Change all `direction` to `travel_flow`
6) "Load a project from a file" at the bottom of the main screen

Or... we could consider doing the backwards compatibility in https://github.com/a-b-street/ltn/blob/629bad71c79a8adbd1e61fc91dfacf2fa2ba7962/backend/src/map_model.rs#L611. I'm more tempted to make temporary / one-off migration tools for the next month, rather than muddy the code with pre-v1 compatibility things though.

XioNoX:
Thanks, that works for me, feel free to close the ticket, up to you if it's worth making it works for less computer savvy users.

Follow up error : https://github.com/a-b-street/ltn/issues/170

PR #206, opened 2025-03-07
Reconcile CNT vs. vanilla LTN landing pages

michaelkirk:
FXIES #205 (part of #104)

This ended up not being a huge change (if you hide whitespace diff). In fact it got much smaller over time as I continued to work on it. But wow, each of these lines was hard fought.

Recall my first attempt was at https://github.com/a-b-street/ltn/issues/104#issuecomment-2658178251 which unified the `ltn` vs `ltn_cnt` landing pages, and introduced a dropdown to switch "app focus" between a "global" view and a "scotland" specific view. 

After talking it through with @dabreegster, it seems like that's not actually likely to reflect how cnt people will use the app. 

Where I landed, conceptually, is very similar from a datamodel perspective! We maintain an "appFocus" but now it's simply hardcoded by the root component based on the URL you access, e.g.: `App(element, {appFocus: "cnt"})`

I could imagine a more dynamic approach, but for now we only have two flavors, so I think it's fine to keep separate pages for `ltn/index.html` and `ltn/cnt.html`

Additionally, you are no longer "redirected" after creating/picking your ltn_cnt project. You stay rooted at `ltn/cnt.html` the entire time.

And as a bonus, the CNT landing page is now rendered in our typical "layout", which I think is nice from a consistency standpoint. 

https://github.com/user-attachments/assets/1691a762-24df-4111-937f-f732e4cad9a4

I think there is further followup we could do. In particular, from #104 

> Should all OSM areas be namespaced, so the URL logic can be less custom?
> Should all savefiles be namespaced as ltn_tool/boundary_name/filename?

Unifying some of these things will let us further de-dupe our implementations that currently have some copy/pasta, but I think they can be done in a subsequent PR.

Special request: I've tested this a bit, but probably not thoroughly enough yet. Can you give it a thorough hammering? I've learned I need to be more careful when mucking with the persistence layer, since we already have some production users.

### DIFF
--- a/web/src/stores.ts
+++ b/web/src/stores.ts
@@ -61,7 +61,7 @@ export type Mode =
 export let map: Writable<Map | null> = writable(null);
 
 // The exact key in local storage
-export let projectName: Writable<string> = writable("");
+export let currentProjectKey: Writable<string> = writable("");
 // False until user activates
 export let showAbout: Writable<boolean> = writable(false);
 
@@ -85,7 +85,7 @@ export let roadStyle: Writable<"shortcuts" | "cells" | "edits" | "speeds"> =
 export let thickRoadsForShortcuts = writable(false);
 
 export function autosave() {
-  let key = get(projectName);
+  let key = get(currentProjectKey);
   if (!key) {
     window.alert("Autosave failed; no projectName set?!");
   }
--- a/web/src/title/NewProjectMode.svelte
+++ b/web/src/title/NewProjectMode.svelte
@@ -10,9 +10,9 @@
     assetUrl,
     autosave,
     backend,
+    currentProjectKey,
     map,
     mode,
-    projectName,
   } from "../stores";
   import { Backend } from "../wasm";
   import { afterProjectLoaded, loadFromLocalStorage } from "./loader";
@@ -37,7 +37,7 @@
         e.detail.boundary,
         undefined,
       );
-      $projectName = `ltn_${newProjectName}`;
+      $currentProjectKey = `ltn_${newProjectName}`;
       afterProjectLoaded();
       // No savefile to load. Create it immediately with just the boundary
       autosave();
--- a/web/src/title/NewProjectMode.svelte
+++ b/web/src/title/NewProjectMode.svelte
@@ -52,6 +52,8 @@
       return;
     }
 
+    // REIVEW: can we change to be like this to align with ltn_cnt project keys?
+    //let key = `ltn/${example}/${newProjectName}`;
     let key = `ltn_${newProjectName}`;
     window.localStorage.setItem(
       key,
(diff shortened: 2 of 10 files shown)


web/src/stores.ts
```diff
@@ -61,7 +61,7 @@ export type Mode =
 export let map: Writable<Map | null> = writable(null);
 
 // The exact key in local storage
-export let projectName: Writable<string> = writable("");
+export let currentProjectKey: Writable<string> = writable("");
```
michaelkirk, line 64: This was just a rename to make it more obvious that it's not *only* the project name.

web/src/title/NewProjectMode.svelte
```diff
@@ -52,6 +52,8 @@
       return;
     }
 
+    // REIVEW: can we change to be like this to align with ltn_cnt project keys?
```
dabreegster, line 55: This would affect newly created projects. But in `getProjectList()`, we still have the old format. We could look for both there, or decide to make a breaking change and switch to `ltn/`. I think we _should_ do that breaking cutover for consistency, but maybe not just yet? Related to #195 also.

web/src/title/loader.ts
```diff
@@ -9,17 +9,44 @@ import { routeTool } from "../common/draw_area/stores";
...
   map,
   mode,
   one_destination,
-  projectName,
   route_pt_a,
   route_pt_b,
 } from "../stores";
 import { Backend } from "../wasm";
 
+// Returns whether the project name is already taken, otherwise the project is created.
```
dabreegster, line 21: I see, this is only used for CNT right now, but we could cut over to it in NewProjectMode too.

dabreegster (approved):
This is very elegant and works great! I love that it functions as a full SPA, instead of being split over two pages. My tests of it all went fine, both for CNT and the other mode. Switching projects and creating new ones all works.

When clicking "Choose project", it was initially unusual to not have the map zoom not reset, but I think I prefer it that way, having gotten used to it.

I vote we defer the question of migrating the non-CNT path to using new-style keys. We do want to do it, but similar to #169, there may be a few more breaking changes in the next few weeks. A clean break and proper savefile versioning in a few weeks would be great, but would just slow us down in the meantime.

michaelkirk:
> When clicking "Choose project", it was initially unusual to not have the map zoom not reset, but I think I prefer it that way, having gotten used to it.

I confess that wasn't an intentional change. In the back of my mind I knew that something had changed, but in my testing, I'd never zoomed very far in, so it wasn't too disorienting to come back to most of the scotland map.

Of course, immediately after merging, I went to edit a neighbourhood (so was quite zoomed in), and it felt pretty disorienting to be back at the national level and only see a few square kilometers of Scotland. Personally, I think it would be better if we zoomed back out to the initial Scotland bbox. 

Can you speak on your preference for the new behavior I introduced? Maybe there's some middle ground that keeps what you like.

dabreegster:
I was picturing a sudden jump in the map, which felt weird, but you're right...the use case in "choose project" demands a lower zoom. Filed #209

PR #211, opened 2025-03-08
Reset map viewport when returning to Choose project and opening a pro…

### LINKED ISSUE #209: Zoom out after clicking Choose project
A followup to #206. If a user is zoomed in on one neighbourhood, then later presses "Choose project", they're left with a map at an inappropriately high zoom for the action of deciding where in Scotland or in the world to have a project. So instead, we should reset the viewport to either Scotland or everywhere. Maybe it'd be nicer if this wasn't just a sudden jump, and we used `fitBounds(bounds, { duration: 300 })` or something like that.

dabreegster:
…ject

Fixes #209 
[Screencast from 08-03-25 11:23:41.webm](https://github.com/user-attachments/assets/4cfbb120-6404-4e89-84ea-4e1c85935710)

### DIFF
--- a/web/src/stores.ts
+++ b/web/src/stores.ts
@@ -1,5 +1,5 @@
 import type { Feature, Polygon } from "geojson";
-import { LngLat, type Map } from "maplibre-gl";
+import { LngLat, type LngLatBoundsLike, type Map } from "maplibre-gl";
 import { type AreaProps } from "route-snapper-ts";
 import { get, writable, type Writable } from "svelte/store";
 import type { Backend } from "./wasm";
@@ -65,6 +65,8 @@ export let currentProjectKey: Writable<string> = writable("");
 // False until user activates
 export let showAbout: Writable<boolean> = writable(false);
 
+export let appFocus: Writable<"global" | "cnt"> = writable("global");
+
 export let backend: Writable<Backend | null> = writable(null);
 export let route_pt_a: Writable<LngLat> = writable(new LngLat(0, 0));
 export let route_pt_b: Writable<LngLat> = writable(new LngLat(0, 0));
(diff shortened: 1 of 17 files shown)


web/src/stores.ts
```diff
@@ -65,6 +65,8 @@ export let currentProjectKey: Writable<string> = writable("");
 // False until user activates
 export let showAbout: Writable<boolean> = writable(false);
 
+export let appFocus: Writable<"global" | "cnt"> = writable("global");
```
dabreegster, line 68: Context is hard to access outside of a component, so I moved it to a store. If you prefer, we could instead try a `ChooseProjectLink` component, so that context is visible

web/src/stores.ts
```diff
@@ -97,3 +99,13 @@ export let useLocalVite: Writable<boolean> = writable(false);
...
 }
+
+export function returnToChooseProject() {
+  mode.set({ mode: "title", firstLoad: false });
+
+  let bounds = [-180, -90, 180, 90] as LngLatBoundsLike;
+  if (get(appFocus) == "cnt") {
+    bounds = [-8.943, 54.631, -0.901, 59.489];
+  }
+  get(map)?.fitBounds(bounds, { duration: 500 });
```
dabreegster, line 110: 500 is maybe a little fast, but anything too long gets annoying. What do you think?

web/src/title/loader.ts
```diff
@@ -154,7 +154,7 @@ export function afterProjectLoaded() {
       writable(0),
     ),
   );
-  get(map)!.fitBounds(get(backend)!.getBounds(), { animate: false });
+  get(map)!.fitBounds(get(backend)!.getBounds(), { duration: 500 });
```
dabreegster, line 157: And the other way: when you open a project, animate the zoom in a bit -- again, quickly.

This zoom happens after the load screen. I wanted to do this _during_ the load screen, but we'd have to mimic the backend `getBounds`, which isn't too clear to do just from the GJ savefile. We could start storing a bbox in there, but I'm not sure it's worth the hassle.

michaelkirk: Doing it *during* loading rather than after would be nice, but I think that could be a low priority followup task.

web/src/stores.ts
michaelkirk: It feels maybe a little fast (definitely not too slow!), but not unreasonable. Let's go with it and we can always tweak it later.

web/src/stores.ts
michaelkirk: A store seems fine!

michaelkirk (approved):
👏

Issue #203, opened 2025-03-06
add schools and community centers to our POI context


michaelkirk:
Related to #157, since we're considering some OSM sourced POIs in our prioritization, we should also add them to our context layers.

- [ ] How should we refer to POIs -- amenities? Land use mix?

dabreegster:
We have schools, GPs, and hospitals already. From the "Prioritization Metrics Methodology' doc, we're missing:

- Greenspaces. There's a (kind of messy) polygon-based dataset from Ordnance Survey we use in NPW.
- Groceries. OSM with some choices of tags should work
- Community centres. OSM!

The doc referenced is https://www.gov.scot/binaries/content/documents/govscot/publications/strategy-plan/2023/02/national-planning-framework-4/documents/national-planning-framework-4-revised-draft/national-planning-framework-4-revised-draft/govscot%3Adocument/national-planning-framework-4.pdf
I'll read through this more closely and iterate on all the categories we need with Cici

michaelkirk:
This is mostly done, and the remainder are tracked in sub-issues, so I'm going to mark this as done!

Issue #157, opened 2025-02-20
prioritization metric: non-residential land use


michaelkirk:
This metric will be a count of POI's within the neighbourhood boundary, divided by the area (POI density).

We will filter which POIs to count, but we won't try to devise any complex differential contribution across the POI types - e.g. a large supermarket will naively be "as valuable" as a little corner store which will be as valuable as a hospital.

Data sources will be hybrid:
- use gov sources where available
  - schools
  - GPs
  - hospitals
- use OSM to fill in the gaps
  - Grocery stores
  - Community centers

I would expect it to make no difference if we buffer the geometry like we did for #156, since AFAIK these POIs won't be "in" the street. So I suppose do whatever's easiest.

## Original Issue description

I'm actually a little fuzzy on this one. 

I understood that we are going to pursue an easily explainable POI *density* — simply count the POIs and divide by area, and not try to calculate some more nuanced/subjective metric for how "well mixed" the land use is.

I am not yet sure what our POI source is, and what subset of those we want to count.

My guess is OSM POIs having some set of reasonable tags, but maybe @dabreegster knows better.

Like #156, we might have to make considerations about buffering our geometries to account for numerical precision.

dabreegster:
So interestingly, the POIs might come from official Scottish gov sources. See schools, GPs/hospitals, railway stations from https://github.com/a-b-street/ltn/blob/main/data_prep/scotland/get_reference_layers.sh. Retail isn't in there; maybe that'll be from OSM, or maybe there's a gov source I'll find via the NPT project.

Greenspace is a little funnier. There's a gov source for this, but it's a bit messy (it has separate polygons inside larger polygons) and definitely isn't a point.

michaelkirk:
I updated the description based on our conversation today.

michaelkirk:
Remaining work seems like a dupe of #203  (tracked in #232, #233)

PR #212, opened 2025-03-08
Add some OSM POIs and start a POI density metric

dabreegster:
Towards #203 and #157.

1) Move the 3 existing POI layers from Scotland-wide GJ files just for display into per study area context data
2) Add two more categories, filled out as we parse the OSM data
3) Start a prioritization metric for POI density

The density scale likely needs tuning like stats19, though Edinburgh looks OK:
image

The raw POIs:
image


backend/src/boundary_stats.rs
```diff
@@ -210,5 +224,22 @@ mod tests {
...
+
+#[derive(Clone, Debug, Deserialize, Serialize)]
+pub struct POI {
+    pub point: Point,
+    pub kind: POIKind,
+    pub name: Option<String>,
+}
+
+#[derive(Clone, Copy, Debug, Deserialize, Serialize)]
+pub enum POIKind {
```
dabreegster, line 239: I'll work with Cici to refine these categories. We may want. Maybe GPs and hospitals should be combined. The only purpose of distinguishing is for the layer, not the metric.

backend/src/create.rs
```diff
@@ -481,3 +496,23 @@ fn nodes_to_edges(graph: &Graph, nodes: Vec<utils::osm2graph::IntersectionID>) -
     }
     edges
 }
+
+fn get_poi(tags: &Tags, point: Coord) -> Option<POI> {
+    if tags.is_any("shop", vec!["convenience", "grocery", "supermarket"]) {
```
dabreegster, line 501: These seem to cover most, but there might be niche things like fruit stands tagged differently. We can validate with local knowledge and iterate.

backend/src/create.rs
```diff
@@ -481,3 +496,23 @@ fn nodes_to_edges(graph: &Graph, nodes: Vec<utils::osm2graph::IntersectionID>) -
...
+        return Some(POI {
+            point: point.into(),
+            kind: POIKind::Grocery,
+            name: tags.get("name").cloned(),
+        });
+    }
+
+    if tags.is_any("amenity", vec!["community_centre", "library"]) {
```
dabreegster, line 509: I'm taking the liberty of lumping libraries into this category too, but may change after discussing with Cici

data_prep/scotland/src/bin/generate_prioritization.rs
```diff
@@ -84,8 +92,7 @@ struct PopulationZoneInput {
 impl PopulationZoneInput {
     fn read_all_from_file() -> Result<Vec<Self>> {
         let population_zones = geojson::de::deserialize_feature_collection_str_to_vec(
-            &std::fs::read_to_string("tmp/population.geojson")
-                .expect("missing population.geojson - see get_reference_layers.sh"),
+            &fs_err::read_to_string("tmp/population.geojson")?,
```
dabreegster, line 95: This crate generates nice error messages with the filename that's missing

data_prep/scotland/src/bin/generate_prioritization.rs
```diff
@@ -115,8 +122,37 @@ struct Stats19Input {
...
+            ("tmp/schools.geojson", POIKind::School),
+        ] {
+            let input: Vec<InputPOI> = geojson::de::deserialize_feature_collection_str_to_vec(
+                &fs_err::read_to_string(path)?,
+            )?;
+            for poi in input {
+                pois.push(POI {
+                    kind,
+                    point: poi.geometry,
+                    name: poi.name,
```
dabreegster, line 152: Nit: We used to show the number of pupils per school. Removed here because I'm not sure it actually matters. We could form a description string instead of name if it is important.

web/src/common/colors.ts
```diff
@@ -33,6 +33,9 @@ export let areaColorScale = commonQuintileColorScale.toReversed();
 export let stats19Limits = [0, 1.0, 10.0, 50.0, 100.0, 1000.0];
 export let stats19ColorScale = commonQuintileColorScale.toReversed();
 
+export let poiLimits = [0, 1.0, 10.0, 50.0, 100.0, 1000.0];
```
dabreegster, line 36: Blindly copying stats19. It seems reasonable enough to start.

dabreegster (commented):
Right before merging, I'll update the Cloudflare context data files

dabreegster:
I'm on battery so won't re-run benchmarks till later. I don't expect checking for more points-in-polygon should be that much of a problem.

web/src/context/POIs.svelte
```diff
@@ -1,17 +1,20 @@
...
   let colors = {
-    Schools: "#7fc97f",
-    GPs: "#beaed4",
-    Hospitals: "#fdc086",
+    School: "#7fc97f",
+    GP: "#beaed4",
+    Hospital: "#fdc086",
+    Grocery: "#386cb0",
+    CommunityCenter: "#ffff99",
   };
```
michaelkirk, line 18: I think using the existing linear-legend is fine for now, because it's expedient. But it strikes me as a little off to have different types of places "on a spectrum", and it's simply not enough space to fit all our POI text in a single row:

Screenshot 2025-03-08 at 13 22 10


To me, something like this makes more sense for categorical data:
Screenshot 2025-03-08 at 12 59 46

(not necessarily the icons, though that'd be cute, I primarily mean the multiline layout: 

```
colored dot: School, colored dot: GP, colored dot: Hospital
colored dot: Grocery, colored dot: Community Center
```

I'm totally fine to merge this as-is for now, but I think addressing this will be necessary in the nearish term.

michaelkirk (approved):
You got farther with this than I was expecting, great!

I pushed up a commit to your branch that I think is uncontroversial. LMK if you'd prefer I don't do that in the future - it was expedient.

backend/src/create.rs
```diff
@@ -61,6 +64,13 @@ impl OsmReader for Osm {
                 node_ids.into_iter().map(|n| node_mapping[&n]).collect(),
             ));
         }
+
+        self.pois.extend(get_poi(
+            &tags,
+            // TODO If a POI is tagged on the building itself, ideally we'd use its centroid. But
+            // an arbitrary point on the boundary is good enough.
+            node_mapping[node_ids.into_iter().next().unwrap()],
```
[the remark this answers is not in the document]
dabreegster: Hah, oops

web/src/App.svelte
```diff
@@ -226,7 +223,12 @@
           {:else if $mode.mode == "new-project"}
             <NewProjectMode />
           {/if}
+
           {#if $backend}
+            {#if $currentProjectKey.startsWith("ltn_cnt/")}
```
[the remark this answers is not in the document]
dabreegster: Done

web/src/PickNeighbourhoodMode.svelte
```diff
@@ -324,6 +332,9 @@
           {:else if selectedPrioritization == "stats19"}
             <b>Pedestrian and cyclist collisions:</b>
             {(props.number_stats19_collisions / props.area_km2).toFixed(1)} / km²
+          {:else if selectedPrioritization == "pois"}
+            <b>Points of interest:</b>
```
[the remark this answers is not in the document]
dabreegster: "Land use density" was also mentioned... will bring up in the next meeting.

web/src/context/POIs.svelte
dabreegster: Oh yeah, I actually meant to comment on running out of space. I'll add a TODO in the issue.

dabreegster:
Thanks for the review, and totally fine to push commits to the branch -- thanks for the better errors.

I'll upload the new files and merge

Issue #207, opened 2025-03-07
chore: migrate ltn (non-cnt) project keys to use the same format as ltn_cnt project keys.


michaelkirk:
As I understand it we currently have exactly 2 key formats:

- For a CNT project it's always: `ltn_cnt/${studyAreaName}/${projectName}`
- For global projects I think it's just `ltn_${projectName}` 

To verify before working on this:

- Does the studyAreaName makes its way into this key for vanilla LTN projects when picking a pre-existing study area?
- Do projects fetched from overpass (not "example" projects) have any special format for keys?

It would simplify our code to have a uniform format because we parse out the substructure of the key. e.g. to determine if it's a CNT project, or to determine which study area it's in.

dabreegster:
> Does the studyAreaName makes its way into this key for vanilla LTN projects when picking a pre-existing study area?

No, but it could be useful to.

> Do projects fetched from overpass (not "example" projects) have any special format for keys?

No, but ditto.


Currently to distinguish the custom Overpass cases, we read the GJ file and look for the `study_area_name` foreign member. I think to decide the map file to load, that field is a good idea. We could make the local storage keys include the study area too, but if you're sending a file to somebody, there won't be a key, and trying to rely on a filename we set by default is brittle (somebody could just rename).

A second thought/concern before making this change: for CNT, the study area names are well-defined and not going to change.. (...Actually, they're Local Authority Districts from the most recent year, but the boundaries**do** actually change over time. A long-term maintenance question could be about how to handle that. I don't think it's very important.)

The global study areas, on the other hand, are very arbitrary things I've drawn. And because they're global, a "flat" namespace for them feels wrong. In A/B Street, we had "country code / city-ish or region / specific study area name". I'm reusing these global study areas for a few other projects, which could be convenient to continue. I'm not sure there's a clear approach how to namespace or even "version" these.

michaelkirk:
> but if you're sending a file to somebody, there won't be a key, and trying to rely on a filename we set by default is brittle (somebody could just rename).

Oh man, I hadn't considered exported projects at all. 

My new theory of everything: The id should be a completely meaningless identifier, and any important semantic information (is it cnt, etc.) should be explicitly encoded into the geojson. 

I don't know that addressing this needs to be top priority or anything, but does that seem like a reasonable place to end up to you?

dabreegster:
[...]

PR #213, opened 2025-03-08
Load saved CNT projects

### LINKED ISSUE #195: Error loading saved project

dabreegster:
Fixes #195 

~~Diffbased on #211, please review that first.~~

Now both the global and CNT "choose project" pages have a tool to load a saved GJ file. From either of those pages, you can load any kind of file, and the app will switch to global or CNT mode accordingly... except for changing the URL.

### DIFF
--- a/web/src/title/LoadSavedProject.svelte
+++ b/web/src/title/LoadSavedProject.svelte
@@ -1,0 +1,40 @@
+<script lang="ts">
+  import { stripPrefix, stripSuffix } from "../common";
+  import { loadFromLocalStorage } from "./loader";
+
+  export let loading: string;
+
+  let fileInput: HTMLInputElement;
+
+  async function loadFile(e: Event) {
+    let filename = fileInput.files![0].name;
+    loading = `Loading from file ${filename}`;
+
+    let contents = await fileInput.files![0].text();
+    let gj = JSON.parse(contents);
+    // Is this a CNT project or a regular one?
+    let key = "";
+    if (gj.study_area_name && gj.study_area_name.startsWith("LAD_")) {
+      let kind = "ltn_cnt";
+      // Parse the project name from the filename, best effort. The user may
+      // have renamed the file.
+      let projectName = stripSuffix(
+        stripPrefix(filename, `${kind}_${gj.study_area_name}_`),
+        ".geojson",
+      );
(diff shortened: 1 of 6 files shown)


web/src/title/LoadSavedProject.svelte
```diff
@@ -0,0 +1,40 @@
...
+
+  async function loadFile(e: Event) {
+    let filename = fileInput.files![0].name;
+    loading = `Loading from file ${filename}`;
+
+    let contents = await fileInput.files![0].text();
+    let gj = JSON.parse(contents);
+    // Is this a CNT project or a regular one?
+    let key = "";
+    if (gj.study_area_name && gj.study_area_name.startsWith("LAD_")) {
```
dabreegster, line 17: This is... brittle and gross. I think part of #207 should also consider adding in appFocus or similar to the GJ savefile itself.

web/src/title/loader.ts
```diff
@@ -47,12 +48,16 @@ export async function createNewProject(
 
 export async function loadFromLocalStorage(key: string) {
   currentProjectKey.set(key);
+  let isCnt = key.startsWith("ltn_cnt/");
+  // TODO Should we also change the URL?
```
dabreegster, line 52: Should we rewrite history and set index.html or cnt.html?

michaelkirk (approved):
LGTM - if you can easily add the url update/redirect, I'd think you should before merging.

If it's challenging, well... I guess this is still a useful improvement as is and we should merge.

web/src/title/loader.ts
dabreegster: I'll get to this tomorrow; I'll merge but keep #195 open.

I like the `cnt/` subdirectory vs `cnt.html`, do we want to go with that too? Should be a pretty easy change in vite config and making a new cnt/index.html.

dabreegster: (It's not challenging I don't think, just late / open PRs weigh on my mind a bit)

michaelkirk: > I like the cnt/ subdirectory vs cnt.html, do we want to go with that too?

Sure! 

TBH in my mind they are semantically equivalent, except that the `.html` communicates that your website has been online since the Clinton administration.

PR #215, opened 2025-03-09
Work on predict impact mode's visualization.

dabreegster:
- Workaround a bug where there are no example routes crossing somewhere
- Add a legend and outline the colors on the map
- Show neighbourhoods, for context where to expect green interior roads

Towards #36. Nothing in this mode here is really that nice yet, these're just little steps.

Before:
image
After:
image

---

I'm also thinking about optionally hiding or de-emphasizing some icons when zoomed. I'm having trouble seeing what's "important" here that the user did to cause all the changes shown:
image
If I hide unedited filters and TRs, the patterns of green and red in the neighbourhood make more intuitive sense:
image
Any opinions here?

### DIFF
--- a/web/src/ImpactDetailMode.svelte
+++ b/web/src/ImpactDetailMode.svelte
@@ -10,10 +10,18 @@
   export let road: Feature;
 
   // TODO Weird to modify the input like this?
-  road.properties!.kind = "focus";
+  let props = road.properties!;
+  props.kind = "focus";
 
-  let routes = $backend!.getImpactsOnRoad(road.properties!.id);
+  let routes = $backend!.getImpactsOnRoad(props.id);
   let idx = 0;
+
+  if (routes.length == 0) {
+    window.alert(
+      "No routes over this road change. (This is a bug in progress of being fixed.)",
+    );
+    $mode = { mode: "predict-impact" };
+  }
 
   function gj(idx: number): FeatureCollection {
     return {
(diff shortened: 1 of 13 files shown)


web/src/ImpactDetailMode.svelte
```diff
@@ -10,11 +10,19 @@
...
-  road.properties!.kind = "focus";
+  let props = road.properties!;
+  props.kind = "focus";
 
-  let routes = $backend!.getImpactsOnRoad(road.properties!.id);
+  let routes = $backend!.getImpactsOnRoad(props.id);
   let idx = 0;
 
+  if (routes.length == 0) {
+    window.alert(
```
dabreegster, line 20: Why is this happening? The count of trips crossing a road before and after edits is changing, but we can't produce any specific examples?
https://github.com/a-b-street/ltn/blob/d1e32b0846c7b9fbf352bdab60fc726f4bc3b345/backend/src/impact.rs#L93
Right now we're insisting both the path before and after edits exists. When one of them doesn't exist, the count will be lower, but we don't return it yet.

I think it's probably better to show one of the paths is missing, similar to #181. I am not sure what's going on yet, but I suspect the route happens to start or end right on a road with a filter, and so it outright gives up instead of going the other direction on the road. So both parts of #181 may actually be pretty important to get better results here!

web/src/PredictImpactMode.svelte
```diff
@@ -54,12 +55,44 @@
...
 
   <div slot="map">
-    <GeoJSON {data} generateId>
+    <GeoJSON data={neighbourhoods}>
+      <FillLayer
+        {...layerId("neighbourhood-boundaries", false)}
+        filter={["==", ["get", "kind"], "boundary"]}
+        paint={{
+          "fill-color": "grey",
+          "fill-opacity": 0.5,
```
dabreegster, line 74: I tried a few and settled on this. In a large study area, I just want to remember where the edits have happened / are targeting

web/src/PredictImpactMode.svelte
```diff
@@ -54,12 +55,44 @@
...
+    <GeoJSON data={impactGj} generateId>
+      <LineLayer
+        {...layerId("predict-impact-outline")}
+        paint={{
+          "line-width": [
+            "interpolate",
+            ["linear"],
+            ["get", "after"],
+            0,
+            1.5 * minRoadWidth,
```
dabreegster, line 88: Line outlines in maplibre have to be done with a second layer underneath that's thicker

web/src/common/index.ts
dabreegster, line 1: Lots of small diffs due to just refactoring this a little

michaelkirk (approved):
I'm not sure if it's new to this PR, but I'm confused by this:

Screenshot 2025-03-10 at 15 31 21

There are 3 numbers on this bit of interface:

before: 470
after: 27
prev/next: 30


What do the prev/next routes correspond to? It seems close to the "after" route count, but not quite equal.

LGTM - feel free to address any of my comments before merging if they seem reasonable, but I'm not overly worried about them in any case.

web/src/PredictImpactMode.svelte
```diff
@@ -54,12 +55,44 @@
     </p>
     <p>
       Thicker roads have more traffic after edits, relative to the max count for
-      any road: {data.max_count.toLocaleString()}
+      any road: {impactGj.max_count.toLocaleString()}
     </p>
+
+    <SequentialLegendBucketed
```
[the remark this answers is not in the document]
dabreegster: image
Good point, done. Separately, there needs to be some adjustment to line up the labels with the vertical lines, even when the labels can get long

dabreegster:
Thanks for the review!

> What do the prev/next routes correspond to? It seems close to the "after" route count, but not quite equal.

I put a little paragraph beneath temporarily. We have zone-to-zone OD data, saying there might be 100 trips from zone 1 to zone 2. Right now, we only pick one random point in zone 1 and 2, calculate that one route, and multiply the per-road-segment counts by 100. That's for faster processing. I'm going to add controls to let the user wait longer but see more accurate results, and think through how to communicate this better.

Issue #123, opened 2025-02-12
Verify neighbourhood drawing in areas with bridges


dabreegster:
Cowgate in EDB is one example. What is the right thing for the user to do, and can the tool do it easily?

dabreegster:
https://www.scottishconstructionnow.com/uploads/CityCentre%20OptC%281%29.png

![Image](https://github.com/user-attachments/assets/403def06-db95-4b0a-a6b8-24163a170223)
This is the desired boundary. This is a really complex circulation plan use case.

Issue #47, opened 2024-03-11
Specify main traffic roads before drawing LTN?


TFCx:
After a few hours of experience, I would consider a small change in the UX workflow. Right now, it's very difficult to reason on more than a very specific and well-formed district.

In my current work, i'm trying to study a few actual / current rat-runs in the city of Montpellier. Especially, I'm trying to find where do they come from upstream-wise. The city doesn't have districts that are completely "sealed" (true LTN) but I can somehow hack the tool by specifying a larger zone. Then those ratruns are clearly identified and I can study them easier than with classic pen and paper.

![image](https://github.com/dabreegster/ltn/assets/9135500/dcc4f4a8-e0f5-480f-8685-1678c55ae6df)

===============

From what I understand of the current workflow:

A) The **project** should be the whole town / metropolitan area on which the user is currently working. I'm pretty sure that for every street, it should belong to one and only one project (Of course, right now, i'm pretty sure some users specify smaller-than-town projects to limit the number of data you need to store/download... but I don't think it should be normal workflow. The frequently downloaded geojson could be cached, no? Especially if the perimeter of the project could follow the perimeter specified by OSM / areacode).

B) Then **neighbourhoods** are the zones the user wants to be LTN / traffic pockets / islands. The main goal is usually to edit them to limit the traversing traffic of each pocket.

Currently, the project forces you to "early on" decide on the perimeter of a single neighborhood you'll be working. I think there's a missing step between A and B where the user could paint the "main" road arteries (i.e. define the main network of the city). Painting those roads would then help to segment your area into several districts.

I'm joining two examples of city layouts with their main networks.

![image](https://github.com/dabreegster/ltn/assets/9135500/e447e3a3-bbfe-4d85-af2a-5fa370a3cb7a)

![image](https://github.com/dabreegster/ltn/assets/9135500/12ae8c52-0d0a-4dea-b6b9-522dc05ea555)

What are your thoughts ? 😬

dabreegster:
> but I can somehow hack the tool by specifying a larger zone.

This is a fascinating use of the tool I hadn't anticipated! It does make sense, as long as you don't care about any potential trips that either start or end inside the area. Your screenshot of the tool has so much going on, I'm happy to see it pushed to limits. Are things like the cell coloring even working and clear enough there?

[...]

PR #216, opened 2025-03-10
primary vs perimeter

### LINKED ISSUE #176: Get rid of "Include Perimeter roads" toggle, always edit perimeter roads.

michaelkirk:
Fixes #176 

Note: the title of #176 is a bit understated - this isn't just about "hiding a toggle" to streamline the UI. Really this is about refining the domain model. With LTNs we're concerned about the separation of __busy__ roads (`primary_road`) from quieter __neighbourhood__ roads (`interior_road`). 

Because we were using "busy roads" as part of our severance logic (osm tag: `[highway]=primary|secondary|etc.`), the perimeter roads often ended up being busy roads, so they were almost synonymous, but not **always**. @dabreegster [posted some good counter examples](https://github.com/a-b-street/ltn/issues/176#issuecomment-2692177073).

With this PR, we simply use the same severance logic to classify `primary_road`, and use those primary_roads in pretty much the same way we were previously using the perimeter roads.

In a lot of cases the output is pretty much identical (minus the shortcuts/calculation and styling change):

**before:**

Screenshot 2025-03-10 at 13 39 47

**after:**

Screenshot 2025-03-10 at 13 39 51

But things definitely get more interesting when you have a primary road cutting through your LTN:

**before:**

Screenshot 2025-03-10 at 13 35 04

**after:**

Screenshot 2025-03-10 at 13 35 09

**before:**

Screenshot 2025-03-10 at 13 42 05

**after:**

Screenshot 2025-03-10 at 14 41 58

One diversion I took: We'd previously talked about having border intersections only on roads that were both `perimeter` and `primary` - but looking at some actual neighbourhoods, I'm not sure that makes sense. I feel like people will be concerned with traffic from `primary_roads` whether that road is on the perimeter or through the middle of their LTN. Consequently, I opted to define a border intersection as *any* connection between a primary and interior (aka non-primary) road. Certainly it's up for discussion though!

A side benefit is we're no longer using "is_perimeter" for anything in the app. It's kind of an expensive calculation, so if we don't need it, that speeds things up

e.g. 
```
bristol_east/build neighbourhood
                        time:   [3.9877 ms 3.9944 ms 4.0019 ms]                   
                        change: [-47.623% -47.510% -47.390%] (p = 0.00 < 0.05)
                        Performance has improved.
```


backend/benches/build_map_model.rs
```diff
@@ -9,18 +9,14 @@ fn benchmark_build_map_model(c: &mut Criterion) {
...
         // Do the file i/o (reading OSM.xml) outside of the bench loop
         let map_model_builder = neighbourhood.map_model_builder().unwrap();
-        c.bench_function(
-            &format!(
-                "build map_model: {name}",
-                name = neighbourhood.study_area_name
-            ),
-            |b| {
+        c.benchmark_group(neighbourhood.savefile_name)
+            .sample_size(neighbourhood.bench_sample_size())
```
michaelkirk, line 13: Some of the Stasbourg benches were painfully slow, so I've lowered the sample size for them.

michaelkirk:
Something that's clear from the screenshots, but that I should have called out explicitly, the traffic cells for LTN's with interior primary roads are quite different from before - separating cells by the severance roads.

michaelkirk:
> Something that's clear from the screenshots, but that I should have called out explicitly, the traffic cells for LTN's with interior primary roads are quite different from before - separating cells by the severance roads.

Thinking about this a little bit, that feels like a bug. I think cell color is just about communicating connectivity - you can still get between these cells by taking the primary road so they should be the same color. I'm going to revert this to a draft until I fix that.

Feel free to hold off on reviewing until then, or you can take it for a spin if you're curious.

michaelkirk:
> I think cell color is just about communicating connectivity - you can still get between these cells by taking the primary road so they should be the same color.

...actually. I'm kind of spiraling on this. Every time I post a message I realize the opposite thing is true. 😆

dabreegster:
It may be morning before I can review this properly...

- I will try some of the corner cases from the other issue, to see how they turn out
- I wonder if we ought to style the primary roads to make them really stand out, like v1 did

> So are those interior primary roads automatically extracted from OSM data? (like primary/secondary roads tags ?). Or can I also modify them?

Kind of relatedly... _sometimes_ when people draw a boundary with interior perimeter roads, they want to reason about the shortcuts going through them, and place a bus gate or something like that, and see how the traffic will detour. Road classification in the UK (and I'm guessing many places) could've been done a long time ago in different policy contexts, and now there could be a desire to reconsider a "primary road" as actually just being a high street with only buses and delivery traffic.

(I'm not sure how this PR impacts this, just trying to get this idea down.)

backend/src/neighbourhood.rs
```diff
@@ -200,11 +192,11 @@ impl Neighbourhood {
...
         }
 
-        if perimeter_roads.is_empty() {
-            // App breaks without perimeter roads: without perimeter roads, there's only one cell,
+        if primary_roads.is_empty() {
+            // App breaks without primary_roads: without primary roads, there's only one cell,
             // so it counts as disconnected (because it doesn't touch a border intersection), and
             // thus we can't calculate shortcuts through it without those intersections.
-            bail!("No perimeter roads");
+            bail!("No primary roads");
```
dabreegster, line 199: Interesting... so we can't operate on something at all if it's all residential-highway according to OSM. I don't know any use cases for that in the first place, though

web/src/layers/NeighbourhoodRoadLayer.svelte
```diff
@@ -80,7 +80,8 @@
   {...layerId("interior-roads-outlines")}
   filter={["==", ["get", "kind"], "interior_road"]}
   paint={{
-    "line-width": lineWidth($thickRoadsForShortcuts, gj.maxShortcuts, 0),
+    // REVIEW: the interior-roads-outline was previously not visible - was that intentional?
```
dabreegster, line 83: Just opening a comment to remind myself to look at this...

dabreegster (commented):
Code-wise, this looks great. The perf boost is awesome.

I need to think through the implications / questions raised about border intersections for any interaction between a primary and interior road, will do that tomorrow!

backend/src/neighbourhood.rs
michaelkirk: Oh! I meant to call this out - I actually wasn't sure what the original comment was about, but I'm assuming it applies equally to the new primary-road based regime.

michaelkirk:
> So are those interior primary roads automatically extracted from OSM data? (like primary/secondary roads tags ?). Or can I also modify them?

They are indeed based on OSM tags.

Currently you cannot modify them in the LTN app. But I think that's a reasonable idea, and something we could add if we go forward with this direction.

dabreegster:
> Consequently, I opted to define a border intersection as any connection between a primary and interior (aka non-primary) road.

Ahh... imagine drawing your boundary around the entire study area. The fact that cells split on either side of these severances makes sense. You would just be showing "all" of the LTNs everywhere, on "both sides" of a severance.

michaelkirk:
After having it sit in the back of my mind for a couple hours, my current best guess is as @TFCx implied:

1. We leave the severance logic how it is phrased in this PR currently (traffic cells don't cross a primary road).
2. We add a tool to manually mark roads as primary/interior when a local expert either:
  - knows that the primary road isn't actually high traffic
  - is explicitly interested in seeing "shortcuts" on the primary road.

dabreegster:
I think that conclusion sounds good to me too. This PR is a great self-contained step, and we can separately think through the flow for overriding this definition.

I'll try a few things out in the morning and likely merge

dabreegster:
All of the examples I'm looking at are great!

image
https://github.com/a-b-street/ltn/issues/176#issuecomment-2692184034 This example again looks good. I'm temporarily coloring the primary roads grey to see them better. This is a case where the user wants to take St John's Road, the primary north/south going through the middle, and treat it like a local road. We can do that with overrides next.

> imagine drawing your boundary around the entire study area. The fact that cells split on either side of these severances makes sense. You would just be showing "all" of the LTNs everywhere, on "both sides" of a severance.

Going back to this thought, here's an example
image
If somebody wanted to work on several LTNs side-by-side at  time, nothing stops them. It generalizes clearly from one.

Shortcuts go between two primary roads. The perimeter of the drawn boundary is not important at all. This is intuitive to me.
image

Looking again at #47, @TFCx had the right idea a year ago. It took us a while to catch up!

web/src/layers/NeighbourhoodRoadLayer.svelte
dabreegster: Before:
image
After:
image

This was a mistake, the outlines should be visible -- I think they help contrast roads and cells. Good catch

backend/src/neighbourhood.rs
dabreegster: Previously, this was kind of impossible unless the user freehand drew a boundary with no roads, or not even a polygon full of roads, and we didn't see any perimeter roads almost parallel to the boundary.

Now, it's defined by primary roads, so you can't do something like this:
image
I've surfaced the error message better. We can think through if it ever makes sense to allow this -- if the user can override the primary/interior roads, then maybe so.

TFCx:
Well, now i'm blushing :blush:. Everything would have been faster if my english wasn't such an obstacle. :sweat_smile: 

Anyway, all the results are wonderful (and coloring the main arteries in grey (or at least widening the roads) is very intuitive). I can't wait to test it on Montpellier or Vendargues.

Just to be sure for the new logic, you should also test it on the non-plannar examples you had (with bridges and tunnels). :)

dabreegster:
Michael, I pushed a commit with a few hopefully not-controversial fixes. I'll merge this now and file another issue to track some next steps for this. Really really awesome!

> Well, now i'm blushing 😊. Everything would have been faster if my english wasn't such an obstacle.

You explained it quite well originally! I've just been really stuck on some assumptions about what the boundary should look like. Thank you for continuing to push us in the correct direction here.

> Just to be sure for the new logic, you should also test it on the non-plannar examples you had (with bridges and tunnels). :)

I tried out the Bristol example and it behaves reasonably. There's a very crazy case in #123 we'll try later.

web/src/AutoBoundariesMode.svelte
```diff
@@ -108,10 +108,7 @@
         mode: "neighbourhood",
       };
     } catch (err) {
-      console.log("error when setting auto-generated neighborhood", err);
```
michaelkirk, line 112: No problem with this change, but I'm curious - were you able to trigger this? (How?)

dabreegster: image
Yes, just draw a boundary inside a larger residential area, and have no primary roads involved

michaelkirk: Ah! I think this is related to https://github.com/a-b-street/ltn/pull/216#discussion_r1988889104

And I've triggered it by having no primary roads in my hand-drawn neighbourood:
Screenshot 2025-03-11 at 10 20 35

PR #217, opened 2025-03-11
Draggable route markers starts on screen.

### LINKED ISSUE #72: Auto-place route markers in a better spot
Route mode and impact-to-one-place mode both randomly start the draggable markers within the bbox, but it's often out the actual study area boundary slightly, or off-screen when first entering the modes.

michaelkirk:
FIXES #72 

They are persisted, which is convenient when you want to see how your changes are affecting the route, but it's confusing, especially for new users, to see this route off screen.

Another thought I had, literally after I'd done all the work and was typing up the commit message:
Maybe another approach would be to leave the logic as is, and zoom out the bounds to encompass the markers.


dabreegster (approved):
This seems like a nice enough step forwards! Changing the zoom / panning out is a nice second idea if we don't get good feedback on this

PR #222, opened 2025-03-11
Rename "primary road" to "main road"

### LINKED ISSUE #220: Try different styling for the primary roads, both in the neighbourhood view and maybe also on the basemap

### LINKED ISSUE #221: Revisit the term "primary road" -- is there a reason to not prefer "main road"? "Primary" kind of suggests OSM's primary/secondary/tertiary hierarchy, which isn't right.

michaelkirk:
FIXES #221 
FIXES #220

In the context of OSM tags, "primary" has an overly specific connotation (`highway=primary`), but our usage is more broad, also encompassing `highway=seconary|tertiary|etc`


dabreegster (approved):
Thanks for the quick fixes!

PR #223, opened 2025-03-11
cartographic delights

### LINKED ISSUE #198: render building footprint above cell coloring
Especially for local experts, the buildings footprints are useful context for orienting yourself.

Try rendering them above the colored cells. 

@dabreegster noted that for areas with high building density, the buildings could effectively obscure the cell coloring. Opacity of the building layer might help in that case.

michaelkirk:
FIXES #198 and a followup item mentioned in #220

Really, we could merge any subset of these commits. I tried to keep them separable.

Apart from the individual design changes, the big structural change is that I've vendored the maptiler style. 

Alternatively we could continue with the previous pattern of downloading and post-processing the given style, but as our diversions get more complex, it becomes increasingly hard to reason about. There's a reason these style files are declarative! 

I also think that it's totally reasonable that we are starting to develop our own cartography - the needs of our app are  special, and unique from some platonic "general purpose map".

## Screenshots 

**before:**

Screenshot 2025-03-11 at 13 24 22

**after:** "main" roads are gray

Screenshot 2025-03-11 at 13 24 52

**before:**

Screenshot 2025-03-11 at 13 24 32

**after:**  most notable here is the relative building and cell contrast

Screenshot 2025-03-11 at 13 24 58

**before:**

Screenshot 2025-03-11 at 13 24 40

**after:** most notable here is the relative building and cell contrast

Screenshot 2025-03-11 at 13 25 21

### DIFF
--- a/web/assets/map-styles/streets-v2-style.json
+++ b/web/assets/map-styles/streets-v2-style.json
@@ -1,0 +1,8165 @@
+{
+  "version": 8,
+  "id": "streets-v2",
+  "name": "Streets",
+  "sources": {
+    "maptiler_attribution": {
+      "attribution": "<a href=\"https://www.maptiler.com/copyright/\" target=\"_blank\">&copy; MapTiler</a> <a href=\"https://www.openstreetmap.org/copyright\" target=\"_blank\">&copy; OpenStreetMap contributors</a>",
+      "type": "vector"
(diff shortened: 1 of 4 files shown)


michaelkirk:
Added some screenshots (how could I forget!)

PR #224, opened 2025-03-12
unify/style project listing

### LINKED ISSUE #148: Project loading screen empty state is confusing
Another small UI/UX thing that I don't think is urgent, but I didn't want to lose track of it:

When you have no projects, this screen is confusing. 

Image

In particular the purpose of "custom" floating there is not obvious. Looking at the code, I think this is intended to be a heading for some kind of grouping, but I haven't investigated enough to know what it's about.

I think a better UX would be to say something like "No saved projects." in that case.

michaelkirk:
FIXES #148 (and beyond!)

## Global

**before:**

Screenshot 2025-03-12 at 12 51 28

**after:**

I've lifted the styles from the neighbourhood picker list (and tweaked them a bit) and am now using them for the project picker list.

Returning users most likely care more about their existing projects, so I've moved those *before* the "create" flow. For new users (screenshots farther down) the won't have any existing projects, so the "create" flow will be at the top.

Note: I'm no longer displaying CNT projects on the main LTN page. The more I think about it, the more I think we should try to maintain that these are two different apps with two different databases. Other than us developers, I don't imagine there will be much cross over between the two. But if this is controversial, it's easy to revert this part https://github.com/a-b-street/ltn/pull/224/commits/50e6a14438553ded84e216a917d409b14fdc06a4

Screenshot 2025-03-12 at 16 34 58

## cnt

**before:**

Screenshot 2025-03-12 at 12 51 57

**after:**

Same idea - project list is before "new projects". It's more drastic here since the "create flow" is so much bigger.

I also stripped "LAD" from the study area name, so it matches with the name from the "create" flow.

Screenshot 2025-03-12 at 16 34 41

## Empty state before (global)

**before:** inexplicable "load a saved file."

Screenshot 2025-03-12 at 12 38 38

**after:** New project is a button

Screenshot 2025-03-12 at 12 58 54

## Empty state before (CNT)

**before:** inexplicable "Or continue with a previously opened file."

Screenshot 2025-03-12 at 12 39 16

**after:**

Screenshot 2025-03-12 at 12 20 27

### DIFF
--- a/web/src/App.svelte
+++ b/web/src/App.svelte
@@ -328,4 +328,66 @@
     font-weight: bold;
     text-decoration: underline;
   }
+
+  /**
+  * Gloal Typography
+  */
+
+  :global(.left .pico h1) {
+    font-size: 36px;
+    margin: 0 0 10px 0;
+  }
+
+  :global(.left .pico h2) {
+    font-size: 28px;
+    margin: 0 0 8px 0;
+  }
+
+  :global(.left .pico h3) {
+    font-size: 22px;
+    margin: 0 0 6px 0;
+  }
+
+  :global(.left .pico h4) {
+    font-size: 18px;
+    margin: 0 0 4px 0;
+  }
+
+  /**
+  * List navigation re-used for neighbourhood and project list
+  */
+  :global(.left .pico .navigable-list h1),
+  :global(.left .pico .navigable-list h2),
+  :global(.left .pico .navigable-list h3),
+  :global(.left .pico .navigable-list h4) {
+    padding: 0;
+    margin: 0;
+  }
+
+  :global(ul.navigable-list) {
+    padding: 0;
+    margin-bottom: 16px;
+  }
+
+  :global(.navigable-list > li) {
+    list-style: none;
+    margin: 0;
+    padding: 8px 8px;
+  }
+
+  :global(.navigable-list > li:not(:last-child)) {
+    border-bottom: solid #ddd 1px;
+  }
+
+  :global(.left .pico .navigable-list .actionable-cell) {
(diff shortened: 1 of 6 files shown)


web/src/App.svelte
```diff
@@ -259,6 +259,15 @@
 </div>
 
 <style>
+  :global(#app .left) {
+    width: 35%;
+    min-width: 400px;
+    max-width: 700px;
+  }
```
michaelkirk, line 266: Previously we had a little bit wider left column for CNT, I think to accommodate the three column LAD list, which makes sense.

But honestly, I think we could do a little wider in general, especially on narrower screens. 

I've also added some min/max widths that I think offer reasonably consistent feeling. All the particular values are up for debate, but I think having *some* kind of min/max is desirable.

web/src/title/loader.ts
```diff
@@ -141,6 +141,61 @@ async function getInputFiles(
...
+  let studyAreas = new Map();
+  let custom = [];
+  for (let i = 0; i < window.localStorage.length; i++) {
+    let key = window.localStorage.key(i)!;
+    if (key.startsWith("ltn_cnt/")) {
+      if (appFocus != "cnt") {
+        continue;
+      }
+      try {
+        console.log("key", key);
```
dabreegster, line 158:
```suggestion
```

dabreegster (approved):
Very very nice, thank you! I agree keeping CNT and non-CNT projects separate makes sense.

web/src/title/loader.ts
michaelkirk: Oops! Thanks.

PR #227, opened 2025-03-12
gray color should only be used for "main" roads

michaelkirk:
Now that we're coloring primary roads gray, let's avoid coloring other roads gray, so as not to confuse.

**before:**
Screenshot 2025-03-12 at 14 24 27

**after:**
Screenshot 2025-03-12 at 14 24 07


michaelkirk:
I'm gonna merge this, since it's fairly trivial.

PR #242, opened 2025-03-16
Impact and route mode improvements

dabreegster:
A few related changes.

## Route mode

#181 found a case where a route exists before edits, but not after. Previously we just showed the red "before" route, which was very confusing. Now it's hopefully more clear:

[Screencast from 16-03-25 14:07:11.webm](https://github.com/user-attachments/assets/ddc8e9b7-67d2-4d94-b274-9f737d6364d0)

Also, made the A/B dot markers orange instead of grey, since grey is pretty hard to spot and also means "main road" now.

## Impact to one destination mode

Previously, when you hover on a road, you got a wall of text, and the blue/red routes were unlabeled mysteries. Now it's hopefully more clear:
image
Think that time ratio should also move to the sidebar and out of the popup? Or are there other things in this mode that don't feel easy to understand?

## Predict impact mode

Part of #36. Previously, you would click some roads with changed counts and get an error message. Now, it's a bit more clear:

image

Zooming in, the bug is clear -- when the start or end road has a filter, we give up immediately. But we should try routing the other direction away from the filter.
image

And the other change is to optionally hide unedited filters and TRs. Without this, it's pretty hard to intuit about the red and green roads:
image
Better after:
image

Should this setting be available outside of this mode? Only danger is that when editing, the user might forget this setting is enabled and get very confused looking at cells, shortcuts, etc.

### DIFF
--- a/web/src/ImpactDetailMode.svelte
+++ b/web/src/ImpactDetailMode.svelte
@@ -3,6 +3,7 @@
   import { GeoJSON, LineLayer } from "svelte-maplibre";
   import { constructMatchExpression } from "svelte-utils/map";
   import { SplitComponent } from "svelte-utils/top_bar_layout";
+  import BackButton from "./BackButton.svelte";
   import { DotMarker, gjPosition, layerId, Link, PrevNext } from "./common";
   import { ModalFilterLayer } from "./layers";
   import { backend, mode, returnToChooseProject } from "./stores";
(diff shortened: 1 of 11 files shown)


web/src/ImpactDetailMode.svelte
```diff
@@ -49,9 +65,7 @@
   </div>
 
   <div slot="sidebar">
-    <Link on:click={() => ($mode = { mode: "predict-impact" })}>
-      Pick a different road
-    </Link>
+    <BackButton on:click={() => ($mode = { mode: "predict-impact" })} />
```
dabreegster, line 68: So the escape key works

web/src/ImpactDetailMode.svelte
```diff
@@ -68,6 +82,19 @@
     </p>
 
     <PrevNext list={routes} bind:idx />
+
+    {#if routes[idx][0] == null}
+      <p style:color="red">
+        No possible route before changes (
```
dabreegster, line 88: I'll remove the "usually a bug" after fixing this case. There could be legit cases where a user manages to disconnect some roads after changes.

michaelkirk (commented):
This all looks good to me!

> Should this setting be available outside of this mode? Only danger is that when editing, the user might forget this setting is enabled and get very confused looking at cells, shortcuts, etc.

I think it would make sense for this to be in our "map layers" control, always accessible, but we can do that in a followup.

I also have a preference for inverting the semantics. Rather than "Only show new filters" (off by default), have it say "Show pre-existing filters" (enabled by default). I think it's a little simpler to reason about, especially if we put it in the layer controls. Checking a box to remove things is atypical.

That could also be done in a followup though.

michaelkirk (approved):

dabreegster:
Thanks for the review! I moved the filter/TR to the list of layers and inverted as suggested; this indeed makes sense

PR #244, opened 2025-03-17
synthetic db keys

### LINKED ISSUE #207: chore: migrate ltn (non-cnt) project keys to use the same format as ltn_cnt project keys.

### LINKED ISSUE #225: "rename" is somewhat broken
Right now the rename feature works iff you manually remove the "ltn_" prefix

But if you, for example, click edit and then save the new name without editing it, the project disappears from the project list.

I assume this is a consequence of us keeping structured data in the key, and the saved key becoming corrupted somehow.

A couple alternative approaches:

1. Keep storing structured data in the key, but dance more carefully with it so we don't corrupt the structure.
2. remove structured data from the key, and store it explicitly.

https://github.com/user-attachments/assets/568f3b61-0835-4825-854f-281c226cfd36

michaelkirk:
⚠️ Don't load this branch locally before reading the instructions below! There's a data migration that I'd like you to test systematically and safely.  

FIXES #207 
FIXES #225 

# Background 

Project data is stored in localStorage (a simple string -> string key value store).

There's broadly two types of project data:
1. the editing state - e.g. where your modal filters are placed, any custom project boundary
2. project metadata - the name of your project, study area, is it cnt

Previously:
1. The editing state was stored in localStorage as a geojson feature collection string
2. project metadata was encoded in localStorage within the key used to reference the editing state. We had slightly different conventions for ltn vs cnt. For example, for ltn the studyArea was encoded in the editing_state geojson, for cnt it was encoded in the editing_state **and** the key. There were also different formatting conventions for the keys
 
```
ltn_my-legacy-custom area: "GeoJson editing state..."
ltn_cnt/LAD_Some-Study-Area/My-Project-Name: "GeoJson editing state..."
```

This was getting increasingly tricky to reason about as the project metadata grew. Leading to bugs like #225.

# Changes

Now projects keys are more opaque. A project has a ProjectID (uuid), and the localStorage key path is a long keypath that includes that new ProjectID.

```
"ltn/projects/by-id/e3139fdc-3dcb-4b16-a398-776629644b18": "GeoJson editing state"
"ltn/projects/by-id/3871082b-0f54-473f-94b1-8156c5093ed6": "GeoJson editing state"
```

Additional data, like the project name and study area, are stored in a separate "index" object.
```
"ltn/projects/index":  
{projects: {3871082b-0f54-473f-94b1-8156c5093ed6: {projectName: "Dundee", studyAreaName: "LAD_Dundee City", appFocus: "cnt"}, e3139fdc-3dcb-4b16-a398-776629644b18: {projectName: "legacy predefinedd", studyAreaName: "bristol", appFocus: "global"}, 9157c1f1-d131-482b-ab27-63fbeafbd094: {projectName: "legacy custom 2", appFocus: "global"}}}
```

To be honest, it's a little more complicated than I'd like, but nothing simpler occurred to me. I actually considered making it even more complicated using indexdb, but that seemed like maybe *too* much (the api is not synchronous). One thing that occurred to me at the end of this, is there's a small opportunity for race conditions if a user is simultaneously editing projects. In general it should be fine, but it's technically possible to corrupt their index, potentially losing access to a project if two tabs save at the same time. We take some care to avoid this in the common case. Using indexdb (which has a notion of transactions) would be a way to be more rigorous around that.

Other notes:

We no longer add study_area_name to the savefile geojson. Since it's in the index, the app doesn't need it.

We do still need it for export though, so during export, studyAreaName and projectName are added as foreign members in the web layer. Note that previously we didn't explicitly add projectName at all - it was only implicit in the filename.

So in effect, I've separated the "save file" format (whats kept in localstorage) from the "export file" format, since their needs are slightly different.

To be honest, I'm not sure this distinction is worthwhile.  The alternative would be, rather than adding the foreign member data upon export, to plumb all that foreign member data to the rust layer, and redundantly include it in the savefile. It wouldn't be used by the app (which gets that data from the index) except during export.

# Review Instructions

## 1. Create some legacy projects

I'm assuming you already have some projects locally. But if not, make sure you have at least:

1. LTN (global) project from a predefined area (e.g. Bristol)
4. LTN (global) project from a custom drawn area (from overpass)
5. CNT project

## 2. Back up your local storage

(based on https://stackoverflow.com/a/61893189/353178)

On main branch, __before checking out this branch__, visit your localhost app and run this in the console:

```javascript
copy(JSON.stringify(JSON.stringify(localStorage)));
```

(Note: There are TWO calls to JSON.stringify)

This should copy your localStorage to your clipboard. 
Paste that into a `localStorage-pre-storage-migration.json` and save it.

### If you need to restore: 

```javascript
// remove any state you've added since you took your backup e.g. the "current schema" key
localStorage.clear()
let serializedLocalStorage = /*paste stringified JSON from clipboard*/
let data = JSON.parse(serializedLocalStorage);
Object.keys(data).forEach(function (k) {
  localStorage.setItem(k, data[k]);
});
```

Reload the app and you should see everything as it was.

## 3. Test Plan

1. Close all app tabs and kill the server

I'm paranoid about the hot-module-reloading making it hard to reason about what is happening when, so let's be deliberate about when the migration runs in case you want to capture any console output.

2. Switch to this branch
4. Rebuild the wasm and restart the vite server.
6. Open a new tab, pop open the browser console, and visit the app

That's it! Hopefully everything worked. You should see in your console output a log entry for each migrated project. And they should be accessible exactly as before.

Only now you can rename them.

If you see any weird output or projects that didn't migrate, let me know and I can investigate. I might need your localStorage backup in that case to reproduce.


michaelkirk:
Ok! I think this is ready for review @dabreegster.

web/src/title/LoadSavedProject.svelte
```diff
@@ -12,24 +13,52 @@
...
     } else {
-      let projectName = stripSuffix(stripPrefix(filename, "ltn_"), ".geojson");
-      key = `ltn_${projectName}`;
+      projectName = stripSuffix(stripPrefix(filename, "ltn_"), ".geojson");
+    }
+
+    if (gj.study_area_name) {
+      studyAreaName = gj.study_area_name;
     }
-    // TODO Be careful with overwriting files
```
michaelkirk, line 30: We now increment a number when importing a file with an existing name. With the new UUID logic they wouldn't get "clobbered", but for sanity, there is some application logic which ensures names must be unique upon save.

michaelkirk:
> Why not put the project name, study area, and CNT bit as foreign members all of the time in the GJ savefile, and plumb through the opaque data in the Rust layer, and then ditch the index? Aka make savefile and export the same. For the list mode, this makes us read + JSON parse + look for these foreign members, which isn't bad at all for the 5MB limit. No need to duplicate state with an index.

I'll give it a shot. To be clear, this means parsing every project in local storage in order to print the list of projects. I think it'll probably be fine. We're likely already passing around bigger entities elsewhere in the app, e.g. the boundary generation.

dabreegster:
The perf impacts of this will be minimal, I think -- browsers are fast at parsing 5MB of JSON. I've done something like this elsewhere before, I'm pretty sure. But do you think otherwise this is the right approach, to not duplicate this state and make the files be self-contained? In a distant future with cloud hosting, there absolutely would be an index, but it'd be transactional and the entire save/load flow would look pretty different anyway. What should we do in the meantime?

michaelkirk:
I think if we're comfortable keeping the entire database "in memory", then having an "index" or other pre-computed "view" of the data is complicating things with little benefit. I agree we should remove it.

michaelkirk:
Ok, I've removed the notion of a summary "index" and all data is stored as foreign members on the individual project documents.

web/src/common/ProjectStorage.ts
```diff
@@ -0,0 +1,376 @@
...
+      console.debug("Schema is already up to date.");
+      return;
+    }
+
+    if (this.storedSchemaVersion < 1) {
+      console.log(
+        `Migrating storage from version ${this.storedSchemaVersion} to ${latestSchemaVersion}`,
+      );
+      for (const focus of ["cnt", "global"]) {
+        const appFocus = focus as "cnt" | "global";
```
dabreegster, line 60: Oh TS...

web/src/title/TitleMode.svelte
```diff
@@ -35,33 +42,51 @@
...
-      window.localStorage.removeItem(key);
-      projectList = getProjectList($appFocus);
+      studyAreas = $projectStorage.studyAreaProjects();
     }
   }
 
-  async function loadProject(key: string) {
-    loading = `Loading project ${key}`;
-    await loadFromLocalStorage(key);
+  async function loadProjectPrompt(projectID: ProjectID, projectName: string) {
```
dabreegster, line 87: nit: I wasn't sure what "prompt" meant. Is this meant to be "with a loading screen"?

dabreegster (approved):
Working like a charm! The rename flow and improvements for displaying project names are great, and the logic for handling old files is very cleanly done. Thanks for re-working this to not use an index

web/src/common/ProjectStorage.ts
```diff
@@ -0,0 +1,376 @@
+import type { FeatureCollection, Polygon } from "geojson";
+
+export type ProjectID = ReturnType<(typeof crypto)["randomUUID"]>;
+export type StudyAreaName = string | undefined;
+export interface ProjectFeatureCollection extends FeatureCollection<Polygon> {
```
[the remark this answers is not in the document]
michaelkirk: Whoops! In #247

web/src/title/TitleMode.svelte
michaelkirk: Yeah, it's not a great name. I updated in #247

PR #247, opened 2025-03-20
synthetic db keys followups

michaelkirk:
Small non-behavioral changes, following up on your feedback in #244


dabreegster (approved):
Thanks for the quick fixes! You got there before I could

PR #248, opened 2025-03-20
FIX: Layers checkbox getting smooshed sometimes.

michaelkirk:
The "existing modal filters and turn restrictions" label is long, and was causing its checkbox to be smooshed.

**before:**

Screenshot 2025-03-20 at 06 32 27

**after:**

Screenshot 2025-03-20 at 06 33 04


dabreegster (approved):
Ahh, thanks!

PR #251, opened 2025-03-20
Fix project migration for files named 'ltn_ltn_foo.txt'

dabreegster:
TODO: Unit test of the migration


michaelkirk:
I pushed up some tests and re-ran the migration from the saved localStorage you'd sent me. Everything seems to work!

michaelkirk (approved):

dabreegster:
Thank you for doing this so quickly! Merging

PR #258, opened 2025-03-20
UI followups to the "add neighborhood" flow

### LINKED ISSUE #256: "draw area" popups are clumsy
Followup from https://github.com/a-b-street/ltn/pull/246#pullrequestreview-2702087008

> I like the clarity of the popup with controls on waypoints, but if you move the cursor quickly and overlap the popup element, the waypoint dragging gets messed up. Also, the extra node markers cover up the popup:
>
> Image
>
> If there aren't some z-ordering tricks we can get working, then we could also try offsetting the popup from the cursor or some other ideas.

michaelkirk:
Followup to #246 

FIXES #256 

I recommend pulling down  the code and running it to test out. It's somewhat interactive and hard to evaluate in screenshots.

**initially:**

Screenshot 2025-03-20 at 12 03 52

**after choosing one:** we mark "step 1" as complete. You can add more or finish.
Screenshot 2025-03-20 at 12 03 56

**after choosing two+:** we mark "step 2" as complete

Screenshot 2025-03-20 at 12 03 58

**after drawing a valid boundary:** you can continue to adjust or finish.

Screenshot 2025-03-20 at 12 16 02

---
Maybe also resolves:

>  I managed to get partly stuck by clicking pregenerated boundaries and merging them, then switching to manual drawing, then switching back to areas, then back to manual. Manual mode didn't have a "Finish / start over" button. I'll repro more carefully next and/or just fix if it's quick.

By offering better hints and keeping the "start over" button available during draw mode, even when you don't have a valid boundary to clear.

Screenshot 2025-03-20 at 12 03 27
---

### DIFF
--- a/web/src/AddNeighbourhoodMode.svelte
+++ b/web/src/AddNeighbourhoodMode.svelte
@@ -287,51 +287,97 @@
         </button>
       </div>
       {#if addMode == "choose-area"}
-        <h3>1. Click an area to start your neighbourhood</h3>
-        <p>
-          These particular boundaries are suggested by finding roads, railways,
-          and water that form severances.
-        </p>
-        {#if neighbourhoodBoundary}
+        <div
+          class="step"
+          class:active={!neighbourhoodBoundary}
+          class:complete={neighbourhoodBoundary}
+        >
+          <h3>1. Click an area to start your neighbourhood</h3>
+          <p>
+            These particular boundaries are suggested by finding roads,
+            railways, and water that form severances.
+          </p>
+        </div>
+        <div
+          class="step"
+          class:active={neighbourhoodBoundary}
+          class:complete={selectedBoundaries.size > 1}
+        >
           <h3>2. Include additional areas</h3>
           <p>Click any adjacent areas you'd like to add to your boundary.</p>
-        {/if}
+        </div>
       {:else if addMode == "draw-area"}
-        <h3>1. Draw your neighbourhood</h3>
+        <div
+          class="step"
+          class:active={!neighbourhoodBoundary}
+          class:complete={neighbourhoodBoundary}
+        >
+          <h3>1. Draw your neighbourhood</h3>
+        </div>
         <AreaControls
           map={notNull($map)}
           bind:waypoints
           bind:drawnShapeOut={drawnShape}
         />
-        {#if neighbourhoodBoundary}
+        <div
+          class="step"
+          class:active={neighbourhoodBoundary}
+          class:complete={selectedBoundaries.size > 1}
+          style="margin-top: 16px;"
+        >
           <h3>2. Refine your boundary</h3>
           <p>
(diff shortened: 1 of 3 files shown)


web/src/AddNeighbourhoodMode.svelte
```diff
@@ -449,4 +495,27 @@
...
+  .step h3::before {
+    /* some content is necessary for the width to take effect. */
+    content: " ";
+    /* approximate width of widest emoji */
+    width: 24px;
+    display: inline-block;
+    margin-right: 8px;
+  }
+  .step.active h3::before {
+    content: "👉";
```
michaelkirk, line 513: Is it reasonable to use unicode/emoji chars like this on non-mac platforms?

I could explicitly include some icons from lucide, it just makes the logic a little more verbose.

web/src/common/draw_area/AreaControls.svelte
```diff
@@ -269,10 +270,9 @@
     on:mouseleave={() => (hoveringOnMarker = false)}
     on:dragstart={startDraggingWaypoint}
     on:dragend={() => (draggingMarker = false)}
-    zIndex={1}
```
michaelkirk, line 272: Screenshot 2025-03-20 at 12 19 37

As is often the case, I found the bug while taking screenshots... by removing this z-index, the dots no longer appear above the popup, but now the waypoints appear (sometimes? always) behind the gray "potential-waypoint" nodes. 

I'm going to keep looking at this.

michaelkirk: Screenshot 2025-03-20 at 13 11 35

Ok, I think I got all my z-ucks in a row. 🦆🦆🦆

web/src/AddNeighbourhoodMode.svelte
dabreegster: Works fine over here, I don't see why not?

web/src/common/draw_area/AreaControls.svelte
```diff
@@ -357,4 +358,20 @@
...
+    /* This solves two problems: 
+    * 1. The popup is too close to the marker
+    * 2. The popup obscures other points, making it hard to click on them.
+    *    To explain further: The popup remains visible when hovering over the marker **or**, once visible, the popup itself. 
+    *    As generic behavior, that makes sense to facilitate clicking on any interactive content within the popup.
+    *    However, the .edit-waypoint-popup doesn't have any interactive content in it, so we don't need that.
+    *    More important, say you have two nearby points positioned vertically, you start with the bottom point, presenting it's popup which covers the top point.
+    *    By adding this padding, there's a little gap between the marker and the popup, so the bottom popup clears while panning to the top point.
+    */
+    padding-bottom: 16px;
```
dabreegster, line 372: Very clever!

Issue #254, opened 2025-03-20
Generating severance boundaries in Highland is slow


dabreegster:
In the release build, it's taking over 60s for me

dabreegster:
The Highland area is truly insane; it covers so much. Even after removing the 3 `touches_*` calculations for a big speedup, it's taking ages. One of the biggest polygon land masses yields 21,000 boundaries! Some ideas to workaround:

- Precalculate auto boundaries for some or all CNT areas, and download them. Still bring in all the prioritization data and be able to calculate stats for dynamic/merged boundaries, but save this initial calculation
- Revisit why we have a study area this massive. These are just local authority districts, but in practice, LTNs only make sense in a few places. Maybe we can manually come up with some more focused study areas, or find a different official boundary dataset more appropriate.

Related to #158, there are some enormous **2000 square km** areas that result that're empty inside.

![Image](https://github.com/user-attachments/assets/5c6a286e-b703-4eef-a9ec-c3018211560d)

dabreegster:
The exported GJ of the boundaries is 100MB! And looking at it, I'm not sure where the 25,000 features are. area_km2's lowest range is some "approximately zero" stuff -- I suspect we might be winding up with some degenerately tiny areas somewhere.

![Image](https://github.com/user-attachments/assets/8b6d6e38-2a70-462e-8d88-bbf8da3dec9a)

michaelkirk:
I'm happy to dig into this if you want to punt it to me. (Though if you're enjoying yourself, by all means continue!)

michaelkirk:
My early measurements were something like 2minutes.
with #250, it's a bit over a minute.

There are some special cases with Highlands, namely the very many small islands.

A couple things I'm trying:
1. Filter out any implausible islands (e.g. under 16km^2 excludes almost all the islands)
2. Putting all the severances in an rtree, so we only try to split each polygon with the relevant severances
3. As of #250, we're filtering out tiny areas before computing their boundary stats, which saves time. Symmetrically, we could filter out huge boundaries, which likely indicate a rural area where LTN treatments aren't plausible. I'm not sure about this one though - because sometimes one corner of a mostly huge rural area could abut a denser area.

I've kind of just thrown a bunch of things at a wall, and need to clean it up, but I'm now seeing runtimes in the 20s range. It might regress a little as I dial in my filters. I'll pick this back up tomorrow.

michaelkirk:
[...]

PR #259, opened 2025-03-20
remove nbhd holes

### LINKED ISSUE #238: Automatic boundaries have an unexpected gap in Dundee
Image
Near https://www.openstreetmap.org/#map=16/56.46005/-2.97536
Haven't investigated cause yet

### LINKED ISSUE #250: Remove holes from "choose area" boundary preview
A neighborhood boundary is a single Polygon without holes.

When building up your neighborhood boundary definition on the prioritization screen, we render the boundary with holes. Holes aren't super common, but can happen when there are lots of severances, like this spot in Dundee:

Image

I think I understand why there are no predefined areas there, and that logic seems sound enough. My concern is what happens when you select the neighborhoods around the hole:

Image

The holes in the boundary preview are confusing because they are removed once you click "create":

Image

I also suspect this might erroneously impact the stats aggregation we're doing on the prioritization screen.

michaelkirk:
Fixes #250 
Fixes #238 

Also partially (and mostly incidentally!) addresses #254 

I haven't actually tested this on Highlands, but I expect the improvements in the Dundee benchmark to carry over for a sizable win. There is still more to do for #254 though.

I recommend reviewing this commit by commit.


dabreegster:
Thank you, this is great!

Issue #243, opened 2025-03-16
Support routing when start or end road has a modal filter


dabreegster:
When the first or last road in a route has a filter on it, we immediately give up. This is definitely wrong; somebody could drive from the other end of the road.

![Image](https://github.com/user-attachments/assets/635866c1-2c1a-412c-8820-b30243ad62d4)

Priority-wise, this affects the interactive route mode and the predict impact mode (very noticeable). I'd say it's pretty important to fix, but I'm not sure how.

# Option 1: precise positions

Right now, the start and end points of our roads is at the granularity of an entire road. In other projects (like [15m](https://a-b-street.github.io/15m/)), it's a road + distance along it, which could be more precise about going "the other way". I'm not sure it's the best time to introduce that complexity here (there are some fun edge cases with routes that are one or two road segments long, involve loop/multi-edge roads, etc).

# Option 2: just allow filters on the start/end roads

Allow starting or from or ending on either intersection on a filtered road. Problem is, sometimes this won't actually show the actual change happening somewhere, and it'll be confusing for users.

# Option 3: only go one direction on a filtered road

If a filter is 10% distance along the road, then it'd usually be intuitive to only allow entering/exiting that road from `dst_i`. If the filter is 70%, then only from `src_i`. This is a cheat that breaks down as the filter is more in the middle.

dabreegster:
I'm going with option 1. It's the only thing that really makes sense. It's a slight investment, but I've done most of the hard work in the 15m project, just need to adapt code here (and cry a little inside about the duplication)

Issue #129, opened 2025-02-12
Improve perf for predict impact


dabreegster:
Perf:

- [ ] Parallelize
- [ ] Reuse the PathCalculator
- [ ] Pass back road GJ once, and then just counts after that
- [ ] Use a basic loading screen (switch to web workers?)
- [ ] Vec for counts, not HashMap
- [ ] Profiling
- [ ] Dedupe requests upfront
- [ ] Measure how long loading the demand model adds. Maybe async load it?
- [ ] Could speed up the Scotland demand data builder with rstar

michaelkirk:
I agree that we should try to get _at least_ the low hanging performance fruit so that the "accurate" prediction is faster, which you've outlined above.

But we should also come at this from the other angle and see how to increase the accuracy of the "fast" mode while keeping it "fast enough".  E.g. one simple thing that should help is to increase the sample count as much as we can. Going from 1->10 samples might provide 80% (totally made up number) of the benefit while still being snappy.

dabreegster:
Oh right, I forgot about that idea! Since every study area is differently sized with more/less OD data, and since people's computers are differently fast, I wonder if we should almost take an approach of "calculate as many as possible within some deadline"

michaelkirk:
Initial profiling:

<img width="1410" alt="Image" src="https://github.com/user-attachments/assets/0f699151-aaa5-4758-8846-d96ffc4c17b3" />

Some things that we're currently spending a lot of time on:

- making requests, in particular this bit:
```
 for _ in 0..iterations {
                let pt1 = self.zones[zone1.0].random_point(&mut rng);
                let pt2 = self.zones[zone2.0].random_point(&mut rng);
                if let (Some(i1), Some(i2)) = (
                    map.closest_road.nearest_neighbor(&pt1).map(|obj| obj.data),
                    map.closest_road.nearest_neighbor(&pt2).map(|obj| obj.data),
                ) {
                    if i1 != i2 {
                        requests.push((i1, i2, trip_count));
                    }
                }
            }
```

In the above code
`random_point` is expensive because it does a `contains` check on an unprepared `zone.geometry`. Maybe we could prepare that geometry for a speedup.

`nearest_neighbor` is a expensive (I think by it's nature, I don't expect we'll be able to speed up that call much)

Or a more radical approach would be to replace this code: rather than picking a random point and mapping it to a road segment, maybe we can pick road segments directly...

Something like:
1. Bucket all the road segments by zone
[...]

PR #260, opened 2025-03-21
More features for impact prediction

dabreegster:
Parts of #36 and #129.

- Impact prediction now has a loading screen to explain frozenness. If/when we manage to plumb progress info from the backend, this'd be the place to show it
- Link to the debug demand UI, reframing it slightly, since it does seem useful
- Let the user choose to either get a very fast-and-inaccurate prediction, or a slower-but-more-detailed prediction. We either sample one A -> B route from each data zone pair, or use as many points as the OD data says to.

Before, with the sampling, the impact of some changes to the center of Dundee:
image

After, without sampling, the visualization gets EXTREMELY noisy. Trips start and end all over the place, so practically every road sees some kind of change:
image

But actually, many of the changes are not interesting. A road going from 9 trips to 0 is a huge relative change, but in absolute scale, it doesn't matter -- that's basically an empty street. So give the user a filter to ignore a changed road if the before and after count are both below the threshold:
image
This defaults to 500, but I'll work with Cici to refine a better default that signifies "so little traffic per day that it doesn't matter". Maybe it's a different threshold for residential vs main roads.

The correctness of all of these results is still in question due to #243.

### DIFF
--- a/backend/benches/impact.rs
+++ b/backend/benches/impact.rs
@@ -1,0 +1,29 @@
+use backend::test_fixtures::NeighbourhoodFixture;
+use criterion::{criterion_group, criterion_main, Criterion};
+
+fn benchmark_impact(c: &mut Criterion) {
+    for fixture in [NeighbourhoodFixture::DUNDEE] {
+        let mut map = fixture.map_model().unwrap();
+        let fast_sample = false;
+
+        // Do the first calculation (making the requests and counts_before) outside of the benchmark
+        map.rebuild_router(1.0);
+        {
+            let mut impact = map.impact.take().unwrap();
+            impact.recalculate(&map, fast_sample);
+            map.impact = Some(impact);
+        }
+
+        c.bench_function("predict impact", |b| {
+            b.iter(|| {
+                let mut impact = map.impact.take().unwrap();
+                impact.invalidate_after_edits();
+                impact.recalculate(&map, fast_sample);
+                map.impact = Some(impact);
+            })
+        });
+    }
+}
+
+criterion_group!(benches, benchmark_impact);
+criterion_main!(benches);
--- a/backend/benches/impact.rs
+++ b/backend/benches/impact.rs
@@ -14,14 +14,16 @@ fn benchmark_impact(c: &mut Criterion) {
             map.impact = Some(impact);
         }
 
-        c.bench_function("predict impact", |b| {
-            b.iter(|| {
-                let mut impact = map.impact.take().unwrap();
-                impact.invalidate_after_edits();
-                impact.recalculate(&map, fast_sample);
-                map.impact = Some(impact);
-            })
-        });
+        c.benchmark_group(fixture.savefile_name)
+            .sample_size(10)
+            .bench_function("predict impact", |b| {
+                b.iter(|| {
+                    let mut impact = map.impact.take().unwrap();
+                    impact.invalidate_after_edits();
+                    impact.recalculate(&map, fast_sample);
+                    map.impact = Some(impact);
+                })
+            });
     }
 }
 
--- a/backend/src/create.rs
+++ b/backend/src/create.rs
@@ -251,7 +251,7 @@ pub fn create_from_osm(
 
         travel_flows,
 
-        impact: None,
+        impact: Some(Impact::default()),
         demand: None,
 
         undo_stack: Vec::new(),
@@ -261,10 +261,7 @@ pub fn create_from_osm(
     };
     if let Some(mut demand) = demand {
         demand.finish_loading(&map.mercator);
-        map.impact = Some(Impact::new(&map, Some(&demand)));
         map.demand = Some(demand);
-    } else {
-        map.impact = Some(Impact::new(&map, None));
     }
 
     let graph = GraphSubset {
--- a/backend/src/impact.rs
+++ b/backend/src/impact.rs
@@ -61,7 +61,6 @@ impl Impact {
                 requests.len()
             );
             self.counts_before = map.router_before.od_to_counts(requests);
-            self.last_fast_sample = fast_sample;
         }
 
         if self.counts_after.is_empty() || fast_sample != self.last_fast_sample {
@@ -74,8 +73,8 @@ impl Impact {
                 .as_ref()
                 .expect("need to rebuild_router")
                 .od_to_counts(requests);
-            self.last_fast_sample = fast_sample;
         }
+        self.last_fast_sample = fast_sample;
 
         let mut features = Vec::new();
         let mut max_count = 0;
--- a/backend/src/impact.rs
+++ b/backend/src/impact.rs
@@ -9,26 +9,22 @@ use crate::{od, MapModel, RoadID};
 /// see how traffic changes across roads in the whole map. This works by finding the best route
 /// before and after changes for every origin/destination "OD" pairs, then counting routes per
 /// road.
+#[derive(Default)]
 pub struct Impact {
     // (r1, r2, count) -- `count` identical trips from `r1` to `r2`
-    requests: Vec<(RoadID, RoadID, usize)>,
+    sampled_requests: Vec<(RoadID, RoadID, usize)>,
+    all_requests: Vec<(RoadID, RoadID, usize)>,
 
+    // Are the two sets of counts calculated with fast_sample or not?
+    last_fast_sample: bool,
     // TODO Can use Vec for perf
     counts_before: HashMap<RoadID, usize>,
     counts_after: HashMap<RoadID, usize>,
 }
 
 impl Impact {
-    /// Calculates `requests` only
-    pub fn new(map: &MapModel, demand: Option<&od::DemandModel>) -> Self {
-        Self {
-            requests: match demand {
-                Some(demand) => demand.make_requests(map),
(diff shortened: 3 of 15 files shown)


backend/benches/impact.rs
dabreegster, line 1: `cargo bench --bench impact` says it's running 0 tests and ASAP exits. I've compared this carefully to other examples, but I'm lost about what I've done wrong -- please help? I want to set this up before starting #129.

backend/src/create.rs
```diff
@@ -261,10 +261,7 @@ pub fn create_from_osm(
     };
     if let Some(mut demand) = demand {
         demand.finish_loading(&map.mercator);
-        map.impact = Some(Impact::new(&map, Some(&demand)));
```
dabreegster, line 264: I haven't measured this, but it should slightly improve initial load time. Before, we picked random points per OD pair. Now we only do it when the user first goes to predict impact mode.

backend/src/impact.rs
```diff
@@ -9,26 +9,22 @@ use crate::{od, MapModel, RoadID};
 /// see how traffic changes across roads in the whole map. This works by finding the best route
 /// before and after changes for every origin/destination "OD" pairs, then counting routes per
 /// road.
+#[derive(Default)]
 pub struct Impact {
     // (r1, r2, count) -- `count` identical trips from `r1` to `r2`
-    requests: Vec<(RoadID, RoadID, usize)>,
+    sampled_requests: Vec<(RoadID, RoadID, usize)>,
+    all_requests: Vec<(RoadID, RoadID, usize)>,
```
dabreegster, line 16: nit: the usize count is redundant here, it's always 1. If it could help memory, we could `iter().map()` to make the two look the same.

web/src/PredictImpactMode.svelte
```diff
@@ -21,8 +35,23 @@
   function pickRoad(f: Feature) {
     $mode = { mode: "impact-detail", road: f };
   }
+
+  async function recalculate(fastSample: boolean) {
+    loading = "Calculating impact";
+    // Render the loading screen before starting the calculation. Unsure why Svelte tick() or one frame doesn't work.
+    await new Promise((resolve) => {
```
dabreegster, line 42: This is consistently working everywhere, thank you for figuring it out! I dunno why one `tick()` doesn't do the trick.

backend/benches/impact.rs
michaelkirk: I don't understand why, but you also need to add something to Cargo.toml whenever you add a new bench. See the others for an example.

dabreegster: OH. Thank you, done

web/src/wasm.ts
```diff
@@ -315,6 +310,17 @@ export class Backend {
   }
 }
 
+export type Impact = FeatureCollection<
+  LineString,
+  { id: number; before: number; after: number }
+> & { max_count: number };
```
michaelkirk, line 316: Oh nice, I didn't know about this `&` syntax. Typescript is so wild.

web/src/PredictImpactMode.svelte
```diff
@@ -57,6 +94,10 @@
       Thicker roads have more traffic after edits, relative to the max count for
       any road: {impactGj.max_count.toLocaleString()}
     </p>
+    <label>
+      Only show roads with at least this much traffic before or after
+      <input type="number" min={0} bind:value={$minImpactCount} />
```
michaelkirk, line 99: What are the units here - "trips per day"? Might be worth labeling.

web/src/PredictImpactMode.svelte
```diff
@@ -46,8 +75,16 @@
...
       This mode estimates the impact of all your changes on traffic around the
       entire area. It's based on many assumptions and must be interpreted very
-      carefully.
+      carefully. <Link on:click={() => ($mode = { mode: "debug-demand" })}>
+        Explore the origin/destination demand data used
+      </Link>
     </p>
+
+    <label>
+      <input type="checkbox" bind:checked={$fastSample} />
```
michaelkirk, line 84: Not a big deal, but a radio might be better at communicating this as a tradeoff:

- [x] Calculate quickly
- [ ] Calculate more accurately

web/src/ImpactDetailMode.svelte
```diff
@@ -76,24 +95,33 @@
...
+
+      {#if routes[idx].count > 1}
+        <p>
+          The routes are currently sampled, to speed things up. This one sample
+          route represents {routes[idx].count.toLocaleString()} trips between the
+          same points.
+        </p>
+        <i>
+          Note: if these don't sum to the total above, that's likely a known
+          software bug
```
michaelkirk, line 109: Is this about #243 or something else?

web/src/ImpactDetailMode.svelte
```diff
@@ -1,52 +1,71 @@
...
-    return gjPosition(routes[idx][1]!.geometry.coordinates[0]);
+    return gjPosition(routes[idx].after!.geometry.coordinates[0]);
   }
 
   function endPos(idx: number): [number, number] {
     let pts =
-      routes[idx][0] != null
-        ? routes[idx][0].geometry.coordinates
-        : routes[idx][1]!.geometry.coordinates;
+      routes[idx].before != null
```
michaelkirk, line 60: yay for named fields!

web/src/ImpactDetailMode.svelte
```diff
@@ -1,52 +1,71 @@
...
+      requestAnimationFrame(() => {
+        requestAnimationFrame(resolve);
+      });
+    });
+
+    routes = $backend!.getImpactsOnRoad(props.id, $fastSample);
+    loading = "";
+    if (routes.length == 0) {
+      window.alert(
+        "No routes over this road change. (This is a bug in progress of being fixed.)",
```
michaelkirk, line 36: Is this #243 or something else?

backend/src/od.rs
```diff
@@ -31,22 +31,19 @@ impl DemandModel {
...
-        // To speed up the impact calculation, how many specific requests per (zone1, zone2)? If
-        // true, just do one, but weight it by count.
-        let fast_sample = true;
-
         let mut rng = WyRand::new_seed(42);
         let mut requests = Vec::new();
 
         for (zone1, zone2, raw_count) in &self.desire_lines {
+            // To speed up the impact calculation, how many specific requests per (zone1, zone2)? If
+            // true, just do one, but weight it by count.
```
michaelkirk, line 46: I imagine theres a really big difference in both the performance and the accuracy when sampling 1 route vs sampling all the routes.

It might be worth exploring having the "fast_sample" do more than 1, still a small enough number that it feels relatively "instant". I don't think you should do that as part of this PR, just thinking about the future.

backend/src/impact.rs
michaelkirk: I'm not very familiar with this part of the app yet, but plausibly there are duplicated entries in `all_requests`? Maybe we could leverage this usize to dedupe them.

No need to do that in this PR, but it might save more space than just getting rid of the usize.

backend/src/impact.rs
```diff
@@ -37,19 +33,48 @@ impl Impact {
...
+            &self.all_requests
+        };
+
+        if self.counts_before.is_empty() || fast_sample != self.last_fast_sample {
+            info!(
+                "Calculating impacts before edits ({} requests)",
+                requests.len()
+            );
+            self.counts_before = map.router_before.od_to_counts(requests);
+            self.last_fast_sample = fast_sample;
```
michaelkirk, line 64: Will this assignment throw off the check 2 lines down? 

My guess is that self.last_fast_sample should be set at the very end of this method.

michaelkirk (approved):
I had one serious concern about a straight forward logic error when invalidating the "fast_sample" path.  If you address that (or if I'm wrong), all other feedback should be considered optional/informational.

web/src/PredictImpactMode.svelte
dabreegster: Done, though I expect we'll adjust this wording more to de-emphasize exact counts. Maybe this filter in the first place is not quite right.

web/src/PredictImpactMode.svelte
dabreegster: Good point, done

web/src/ImpactDetailMode.svelte
dabreegster: Yes, it's this

web/src/ImpactDetailMode.svelte
dabreegster: Same, yes

backend/src/od.rs
dabreegster: Yep, agreed. Ideally I'd love to expose some kind of way for the user to say how long they're willing to wait, cut things off at that point, and give some sense of the accuracy achieved.

backend/src/impact.rs
dabreegster: AHHH. Good catch, thank you!

backend/src/impact.rs
dabreegster: #129 has this idea, thanks! I'm hoping the benchmark started will capture the effects of incremental improvements like this

PR #263, opened 2025-03-22
Misc style changes

dabreegster:
Just different small things I've noticed. Can take or leave any of these individually.

- Highlight the neighbourhood row when hovering on the sidebar. When you hover on the link text, it styles as hovering, but I got a bit surprised when seeing an area on the map get highlighted when my mouse was on the whitespace left of the edit/delete buttons

- Tune minzoom for neighbourhood things. Things like border arrows were just black outline blobs at some zooms, and having edit popups appear when you're really zoomed out felt odd. Now as you approach an area, you just see cells:
Screenshot from 2025-03-22 12-54-19
Then a bit more zoomed in, everything else appears (this is a really busy example):
Screenshot from 2025-03-22 12-54-24

Remove a dashed city/misc admin border layer from the basemap. In one place I spotted it, it's just noise that's confusing with our other styling. Before:
Screenshot from 2025-03-22 12-44-00
After:
Screenshot from 2025-03-22 12-46-05

And now that we have the "existing filters" layer, show the controls for global mode.


web/src/common/ProjectStorage.ts
```diff
@@ -7,7 +8,7 @@ export interface ProjectFeatureCollection
   // Foreign Members
   project_name: string;
   study_area_name: StudyAreaName;
-  app_focus: "global" | "cnt";
```
michaelkirk, line 10: This type was long overdue. Thanks for cleaning it up!

web/src/common/ProjectStorage.ts
```diff
@@ -1,4 +1,5 @@
 import type { FeatureCollection, MultiPolygon, Polygon } from "geojson";
+import type { AppFocus } from "../stores";
```
michaelkirk, line 2: This is an aside, but I recall hitting problems that I ultimately blamed on circular import references. 

The error was something like "`new Database()` is not a constructor", and my resolution was to make the imports more DAG like.

IIRC importing stores into ProjectStorage was one of the cases that tripped me up before. But this seems to be building fine, so now I'm confused.  Any idea why?

michaelkirk (approved):
All nice improvements!

web/src/common/ProjectStorage.ts
dabreegster: It's a good point generally -- we could maybe have a `type.ts` with no dependencies itself. I suspect this is fine because it's a TS type definition import only?

PR #264, opened 2025-03-22
Only generate boundaries within settlements.

### LINKED ISSUE #158: autogenerated neighbourhood boundary picker should not suggest unlikely neighbourhoods (e.g. a cow field)
Currently our autogenerated boundaries use road topology to slice up as best it can the entire map, which means that it's possible to pick an area that would never be suitable for an LTN.

On the call it was expressed that this could lead to the tool feeling "dumb" or "untrustworthy".

A simple metric floated was to filter out anything below a certain density.

Maybe relevant is that some of the discomfort with the current situation is the bright coloring of everything being a bit overwhelming, which might be less of an issue when we have semantically meaningful (prioritization based) coloring.

It was also mentioned that this work could likely be done *after* remaining feature work without being disruptive — it isn't expected to change any app flow.

### LINKED ISSUE #254: Generating severance boundaries in Highland is slow

michaelkirk:
An attempt at https://github.com/a-b-street/ltn/issues/254#issuecomment-2744699163

FIXES #158 
FIXES #254

**before:** Largely rural areas (Highlands) had many huge, improbably, suggested boundaries, and ~30s load time.

Screenshot 2025-03-25 at 09 18 28

Though it did look kind of fun.

**after:**
Screenshot 2025-03-25 at 09 18 08

Zooming in to Inverness
Screenshot 2025-03-25 at 09 18 15

Denser areas, e.g. Glasgow City are more or less unaffected by this change:

**before:**
Screenshot 2025-03-25 at 09 19 42

**after:** Some of the outskirt neighborhoods seem to have an additional split, presumably when a previously "severance based" boundary is now split by the settlement boundary. I think that's fine:

Screenshot 2025-03-25 at 09 19 25

If we decide to, I think adding buffering would be a straight forward change on top of this. But in my own QA, I haven't yet found cases where it would be helpful.

Performance diff:

```
inverness/generate auto boundaries (and stats)
                        time:   [4.5871 s 4.6247 s 4.6708 s]
                        change: [-79.509% -78.969% -78.414%] (p = 0.00 < 0.05)
                        Performance has improved.
dundee/generate auto boundaries (and stats)
                        time:   [147.92 ms 148.44 ms 149.01 ms]
                        change: [+107.50% +108.33% +109.26%] (p = 0.00 < 0.05)
                        Performance has regressed.
```

The fast ones got quite a bit slower (relatively), but I think are still fast enough. While the slow ones got a **lot** faster. I think this puts us into an acceptable range for Highlands.

### DIFF
--- a/backend/src/auto_boundaries.rs
+++ b/backend/src/auto_boundaries.rs
@@ -2,7 +2,7 @@ use crate::boundary_stats::BoundaryStats;
 use crate::geo_helpers::buffer_polygon;
 use crate::{MapModel, NeighbourhoodBoundary, NeighbourhoodDefinition};
 use anyhow::Result;
-use geo::{Area, BoundingRect, Coord, Intersects, LineString, MultiPolygon, Polygon, Relate};
+use geo::{Area, BoundingRect, Coord, LineString, MultiPolygon, Polygon, Relate};
 use geojson::{Feature, FeatureCollection};
 use i_overlay::core::fill_rule::FillRule;
 use i_overlay::float::slice::FloatSlice;
@@ -28,9 +28,8 @@ impl MapModel {
                 if let Some(context_data) = self.context_data.as_ref() {
                     context_data
                         .settlements
-                        .bounding_rect()
-                        .unwrap()
-                        .intersects(&line_string.bounding_rect().unwrap())
+                        .relate(*line_string)
+                        .is_intersects()
                 } else {
                     true
                 }
@@ -71,6 +70,13 @@ impl MapModel {
             // .0025km (50m x 50m)
             let min_area_km_2 = 0.0025;
             if area_km_2 < min_area_km_2 {
+                continue;
+            }
+
+            // Truly huge areas typically indicate "all the non-settlement" land - a negative left
+            // behind by removing all the rural areas.
+            let max_area_km2 = 50.0;
+            if area_km_2 > max_area_km2 {
                 continue;
             }
             if let Some(context_data) = self.context_data.as_ref() {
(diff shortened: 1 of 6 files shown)


michaelkirk:
One tricky bit are the settlement outskirts. 

So we split the LAD into boundary polygons using severances as the splitters. However many areas on the inner edge of the settlements ("outskirts"), end up in the huge mostly "unsettled" severance polygon. That one, being huge and representing the negative space (mostly) between settlements, gets removed. Meaning, the outskirts are removed as collateral damage.

My attempt at addressing this was to introduce the settlement boundaries themselves as a splitter, along with the traditional "severances".

**splitting by severances only:**

arboroath-no-outskirts

**splitting by severances and settlement boundary:**
arboroath-with-outskirts

michaelkirk:
I did a little cleanup and updated the description. I think this is a big improvement. I think we should merge it and get sustrans to try it out.

backend/src/auto_boundaries.rs
```diff
@@ -23,10 +23,53 @@ impl MapModel {
...
+                    // To get an intuition, imagine a classic ring-road topology. The ring road is typically
+                    // near the edge, but not *at* the edge of a settlement. I'm calling that area
+                    // of the settlement beyond the ring road the "outskirt".
+                    //
+                    // If we were splitting boundaries only between severances, then this outskirt
+                    // will become part of a potentially huge boundary that starts from just
+                    // outside the ring road and continues until it hits another major severance,
+                    // which could be a 100 miles away.
+                    //
+                    // NOTE: This approach is an improvement, but isn't perfect. Especially with small
```
dabreegster, line 51: Yeah, in all honesty, if somebody needs to work on something around the edge of a study area, they're going to have a much better time just importing a custom boundary that includes the right surrounding area. I don't think in practice this will need to happen.

dabreegster (approved):
This makes Inverness actually usable, so yes, let's go for it and get feedback on any edge cases needing to form neighbourhoods right near settlement boundaries.

dabreegster:
I've reuploaded the prioritization data

Issue #218, opened 2025-03-11
Let the user override the definition of primary roads


dabreegster:
Some small followups after #216:

- [ ] Try different styling for the primary roads, both in the neighbourhood view and maybe also on the basemap
- [ ] Revisit the term "primary road" -- is there a reason to not prefer "main road"? "Primary" kind of suggests OSM's primary/secondary/tertiary hierarchy, which isn't right.

And the bigger idea, captured partly by #47 and https://github.com/a-b-street/ltn/pull/216#issuecomment-2711977980:

- [ ] Allow the user to override the definition of primary vs interior roads, when either local knowledge differs from the OSM highway classification, or when the user wants to model an effective change to that classification

michaelkirk:
As the sub-issues are tracked separately, I'm going to close this.

#279 
#278

PR #265, opened 2025-03-22
Let the user change main road classification

dabreegster:
#218 

The simplest start, treating this pretty much the same as toggling travel flow. Not ready for review.

Questions to raise:
- Does it make sense to let someone turn a local road into a main road? Yes, to undo naturally and because why not? Could be cases where valid
- Should we use the changeable definition for autosplitting boundaries? No, makes the ordering confusing; most of the time this will not matter anyway

TODO:
- Better icon (thick roads with thin ones)
- Border arrows in the middle of the neighbourhood cover everything up
- Selecting one road segment at a time is awful. I at least want to select all segments with the same name inside the neighbourhood. Maybe a checkbox to do one-at-a-time vs bulk?


TFCx:
> Does it make sense to let someone turn a local road into a main road? Yes, to undo naturally and because why not? Could be cases where valid

I have at least one example in mind: I'd like to test a proposal to have tight neighbors (LTNs) in Montpellier encircled by a ring of main roads in Montpellier (to mimic Gant's trafic plan). Those main roads already exist on 3/4 of the perimeter but the last 1/4 is a residential zone. I would like to be able to try to create/study a proposal that need some of those roads to turn into main one-way arteries. :)

michaelkirk:
I've done a bit more on this, and I think it makes sense to merge at this point.

https://github.com/user-attachments/assets/f19a85b7-60e5-4322-8e7a-6cd628b76ad2

There is some outstanding work, but I personally think we should leave it for a followup PR - WDYT?:

- [ ] Border arrows in the middle of the neighbourhood cover everything up
  - I see what you mean. I think we should improve it, but it's already an issue, and I don't think being able to reclassify roads necessarily makes this more or less urgent. I've got most of a solution in another branch, but it might be controversial, so I'd kind of like to propose it separtely. The premise is to switching from representing the "border arrows" as features with lat/lon based geometry generated in the backend, to using a symbol layer, so that it can scales with the zoom, getting smaller as you zoom in. 

- [ ] Selecting one road segment at a time is awful. I at least want to select all segments with the same name inside the neighbourhood. Maybe a checkbox to do one-at-a-time vs bulk?

I think this should be improved too. But honestly, in my testing it's not all that bad. It's pretty comparable to the way the "one-way tool" works, right? It might make sense to leverage the same solution in both places.

michaelkirk:
Since you opened the PR, github won't let me request a review, but "please take a look" @dabreegster =)

TFCx:
> There is some outstanding work, but I personally think we should leave it for a followup PR - WDYT?:

<joke> You're right, it is some outstanding work </joke> :sweat_smile:

web/src/edit/NeighbourhoodMode.svelte
```diff
@@ -485,6 +524,8 @@
               </div>
             {:else if action.kind == "oneway"}
               <p>Click to change direction</p>
+            {:else if action.kind == "main-roads"}
+              <p>Click to toggle this a main / minor road</p>
```
dabreegster, line 528: Oops, I was lazy here too, trying some slightly less weird wording

PR #262, opened 2025-03-22
generate-boundaries perf improvements

michaelkirk:
A few difference changes towards #254. You can see the individual commits for more details about how much each change helped.

Even with these improvements, highlands is still too slow I think. I mentioned another idea I'd like to pursue here: https://github.com/a-b-street/ltn/issues/254#issuecomment-2744699163

```
$ cargo bench --bench=boundary_stats -- --baseline=inverness-perf
    Finished `bench` profile [optimized] target(s) in 0.06s
     Running benches/boundary_stats.rs (target/release/deps/boundary_stats-1c26d39854e3ef0d)
Gnuplot not found, using plotters backend
build stats: Longman in LAD_Highland
                        time:   [594.43 µs 594.70 µs 595.00 µs]
                        change: [-8.7287% -8.4349% -8.1537%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 13 outliers among 100 measurements (13.00%)
  2 (2.00%) low mild
  2 (2.00%) high mild
  9 (9.00%) high severe

Benchmarking inverness/generate auto boundaries (and stats): Warming up for 3.0000 s
Warning: Unable to complete 10 samples in 5.0s. You may wish to increase target time to 201.1s.
inverness/generate auto boundaries (and stats)
                        time:   [22.055 s 22.486 s 22.868 s]
                        change: [-61.770% -60.712% -59.689%] (p = 0.00 < 0.05)
                        Performance has improved.

build stats: Hilltown in LAD_Dundee City
                        time:   [840.77 µs 842.52 µs 844.72 µs]
                        change: [-18.688% -18.133% -17.496%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 13 outliers among 100 measurements (13.00%)
  3 (3.00%) low mild
  7 (7.00%) high mild
  3 (3.00%) high severe

Benchmarking dundee/generate auto boundaries (and stats): Warming up for 3.0000 s
Warning: Unable to complete 100 samples in 5.0s. You may wish to increase target time to 7.3s, or reduce sample count to 60.
dundee/generate auto boundaries (and stats)
                        time:   [72.682 ms 72.872 ms 73.063 ms]
                        change: [-4.6683% -3.9062% -3.1746%] (p = 0.00 < 0.05)
                        Performance has improved.

```

PR #267, opened 2025-03-22
Only filter out 'islands' when the input boundary is a multipolygon

dabreegster:
#262 accidentally broke small study areas, treating the one boundary polygon as an island. Fix by just seeing if the study area boundary is actually a multipolygon before checking for small islands.

Tested in the global version prebuilt areas like Montlake and Elephant & Castle


michaelkirk (approved):
Good catch, thank you!

PR #268, opened 2025-03-24
Legend touchups

### LINKED ISSUE #231: Rework crowded layout for POI legend (and other categorical labels?)
We have two broad categories of legends:

"linear" and "categorical"

Linear represents a continuous range of values, like population density. 
Categorical represents discrete unreleted values like POI categories.

## Linear 

Linear legends labels can be drawn "bucketed" or "limited".

### Linear Bucketed (label the colors)

Image

### Linear Limited (label between the colors)

Image

I feel like the linear legends are looking pretty decent now.

## Categorical

A couple of our categorical legends are too crowded, making it unclear which label corresponds to which color

Image

Image

Additionally there are a couple categorical legends which could arguably be linear bucket legends...

Image

Image

I think maybe the history here is that the categorical buckets used to be the way to label the color, rather than between the color. At the time I introduced "bucketed linear" legends, I didn't realize we had this capability. Since introducing the bucketed linear legends though, I've made some improvements to the styling, so I'd like to keep it.

I think our approach to uncrowd the POI and Cycle Infra will be to do something closer to the "vertical" mode of the categorical legend (a feature we're not currently using anywhere AFAICT).

michaelkirk:
Fixes #231 

Addresses some of the crowding issues in the current legend layout.

The legend color swatches are now circles for POI and Collision data, which matches their cartographic symbology.

---
**POI before:** crowded squares

Screenshot 2025-03-24 at 15 25 04

**POI after:** less crowded circles

Screenshot 2025-03-24 at 15 26 07

---
**LoS before:** Labels are under limits

Screenshot 2025-03-24 at 15 26 57

**LoS after:** Labels are under buckets

Screenshot 2025-03-24 at 15 27 44

---

**Cycle infrastructure before:** crowded, label alignment is a little rough

Screenshot 2025-03-24 at 15 28 56

**Cycle infrastructure after:** less crowded

Screenshot 2025-03-24 at 15 29 03

---

**Collisions before:** not bad before (though I'd center the labels)

Screenshot 2025-03-24 at 15 31 40

**Collisions after:** legend symbology matches map (circles). I also reversed the ordering - it seemed more intuitive.

Screenshot 2025-03-24 at 15 32 35

### DIFF
--- a/web/src/common/QualitativeLegend.svelte
+++ b/web/src/common/QualitativeLegend.svelte
@@ -1,0 +1,31 @@
+<script lang="ts">
+  export let labelColors: Record<string, string>;
+  export let swatchClass: "rectangle" | "circle" = "rectangle";
+  export let itemsPerRow: 1 | 2 | 3 | 4 = 2;
+  let gridItemWidth = `${100 / itemsPerRow}%`;
+</script>
+
+<div
+  style="display: grid; grid-template-columns: repeat(auto-fill, calc({gridItemWidth} - 10px)); gap: 10px;"
+>
+  {#each Object.entries(labelColors) as [label, color]}
+    <div style="display: flex; align-items: center; gap: 6px;">
+      <span class="color-swatch {swatchClass}" style:background={color} />
+      {label}
+    </div>
+  {/each}
+</div>
+
+<style>
+  .color-swatch.circle {
+    height: 20px;
+    width: 20px;
+    border-radius: 10px;
+    border: 1px solid black;
+  }
+  .color-swatch.rectangle {
+    height: 16px;
+    width: 24px;
+    border: 1px solid #888;
+  }
+</style>
--- a/web/src/common/SequentialLegend.svelte
+++ b/web/src/common/SequentialLegend.svelte
@@ -3,6 +3,7 @@
   type IntoLabel = string | number;
   export let labels: { limits: IntoLabel[] } | { buckets: IntoLabel[] };
   export let decimalPlaces = 0;
+  export let fullWidthBucketLegend = false;
 
   let intoLabels: IntoLabel[] = [];
   if ("limits" in labels) {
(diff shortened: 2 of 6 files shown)


web/src/common/SequentialLegend.svelte
```diff
@@ -3,6 +3,7 @@
   type IntoLabel = string | number;
   export let labels: { limits: IntoLabel[] } | { buckets: IntoLabel[] };
   export let decimalPlaces = 0;
+  export let fullWidthBucketLegend = false;
```
michaelkirk, line 6: This is kind of esoteric...

Recall that we can choose between labeling the buckets themselves vs. labeling the space between the buckets. 

When labeling the limits, in order to center the label under the limit, we need to add some padding to the colors scale. When labeling the buckets, we technically don't need that padding, but from our previous usage in the prioritization module, I liked having the legend colors the same width regardless of if we were labeling limits vs. buckets, because it minimizes visual jank when switching units.

https://github.com/user-attachments/assets/bb8f9c61-dfc1-489f-887c-527fc9e4808b

However, the map legend is pretty narrow, so that "needless" padding comes at a cost.

**before adding this option:** This padding feels superfluous here
Screenshot 2025-03-24 at 15 18 36

**after this change:** Remove the padding in this case, note it lets the text wrap to two lines, rather than three.

Screenshot 2025-03-24 at 15 18 21

web/src/common/QualitativeLegend.svelte
```diff
@@ -0,0 +1,31 @@
+<script lang="ts">
```
michaelkirk, line 1: This was forked from svelte-utils, but changed significantly. 

I didn't realize it at the time, but when I added the "bucketed" mode to SequentialLegend, I didn't realize that it was more or less functionally redundant to `<QualitativeLegend horizontal />`. I made some other stylistic change, so it wasn't a complete waste, but I might have gone about it differently had I known.

So the stable of labels looks like:

For labeling values on a dimension (e.g. counts):

- `Sequential legend` with "buckets" to center labels on the color bands.
- `Sequential legend` with "limits" to label *between* the color bands

For categorical labels that don't fall on a single dimension:
- `Qualitative Legend`

This is just a "rule of thumb", and somewhat subjective. e.g. the crash severity is pretty easy to think about as a single dimension rather than categories, but I thought it was worthwhile to match the circle symbols on the map.

"level of service" could arguably be categories or not.

dabreegster: You've basically come up with a design system and components for it, this is all great!

dabreegster (approved):

PR #269, opened 2025-03-25
Cleared "one way" button

### LINKED ISSUE #200: UI confusion: road direction vs. turn restriction tools
It's clear we need the ability to do both of these things, but they are somewhat redundant and overlapping. 

Some simple things we can do to clarify them:

## icons 

In particular, the current "reverse direction" icon is vague.

Image

## copy changes

Angus suggested that "one way" is a more recognizable term. Even though it's not *technically* complete (it's also a "2-way" tool!), this might be the kind of helpful lie that makes things more intuitive.

michaelkirk:
Fixes #200 

If I recall correctly, this conversation might have happened during the meeting when you were having network troubles - but we talked about how the "travel flow" tools purpose wasn't immediately obvious.

**before**

Screenshot 2025-03-24 at 17 31 45

**after**

Screenshot 2025-03-24 at 17 31 35

### DIFF
--- a/web/assets/one_way_left.svg
+++ b/web/assets/one_way_left.svg
@@ -1,0 +1,42 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<svg
+   width="200"
+   height="200"
+   viewBox="0 0 208 208"
+   version="1.1"
+   id="svg4"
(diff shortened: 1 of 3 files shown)


dabreegster:
Thanks! I like the new icon

PR #270, opened 2025-03-25
FIX: "pick neighbourhood" shouldn't look disabled

### LINKED ISSUE #237: neighbourhoods on PickNeighbourhoods look disabled because they are gray
If you have no prioritization metric picked, the neighbourhoods that you can click on are colored gray, a design which idiomatically implies "disabled". We should style them differently to make it obvious to the user that they are supposed to click on one of them.

Image

michaelkirk:
FIXES #237

**before:** On the pick neighbourhood screen currently, the pickable areas are colored gray, which communicates "disabled". E.g. we use the same gray color as a mask to "blot out" the rest of the world outside of the study area.

Screenshot 2025-03-25 at 11 43 18

**after:**

https://github.com/user-attachments/assets/375f7795-ca20-4464-bd7b-8e0ee269b3c8

Now we use the same boundary outline style as used when building the neighbourhood boundary.

I think it's nice to throw back to the same design we use elsewhere to communicate neighbourhood boundaries.

I also think it's nice to match the "highlight" color on the map with the color we use in the sidebar when hovering.

I also like that the lack of mask when *not* hovered means you can see the interior of the neighbourhoods.

The major downside to this approach is that without a visible "fill", it might not be obvious enough that the neighbourhoods represent a 2-d area. 

I could easily be convinced to add a low-opacity fill to the unhovered state. It shouldn't be gray though. Maybe blue like we show for the LAD boundaries on the CNT title page?

Or maybe it's fine as phrased. What do you think?

### DIFF
--- a/web/src/PickNeighbourhoodMode.svelte
+++ b/web/src/PickNeighbourhoodMode.svelte
@@ -1,7 +1,7 @@
 <script lang="ts">
   import type { FeatureCollection } from "geojson";
   import { CirclePlus, Pencil, Trash2 } from "lucide-svelte";
-  import type { DataDrivenPropertyValueSpecification } from "maplibre-gl";
+  import { type DataDrivenPropertyValueSpecification } from "maplibre-gl";
   import {
     FillLayer,
     GeoJSON,
@@ -12,7 +12,7 @@
   import { downloadGeneratedFile, notNull } from "svelte-utils";
   import { Popup } from "svelte-utils/map";
   import { SplitComponent } from "svelte-utils/top_bar_layout";
-  import { HelpButton, layerId, Link } from "./common";
+  import { HelpButton, layerId, Link, Style } from "./common";
   import { pickNeighbourhoodName } from "./common/pick_names";
   import { ModalFilterLayer } from "./layers";
   import {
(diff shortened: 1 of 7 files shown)


web/src/PickNeighbourhoodMode.svelte
```diff
@@ -125,15 +126,22 @@
...
   ): DataDrivenPropertyValueSpecification<number> {
-    return {
-      none: hoverStateFilter(0.3, 0.5),
+    let highlightedOpacity = 0.5;
+    let styles: Record<string, DataDrivenPropertyValueSpecification<number>> = {
+      none: [
+        "case",
+        ["==", ["feature-state", "highlight"], "yes"],
+        highlightedOpacity,
+        hoverStateFilter(0.0, highlightedOpacity),
```
dabreegster, line 135: I like a very slight fill -- 0.2 here feels good:
image
Yellow looks OK, but blue would look fine too. But yeah, I think somehow styling as a 2D area is still important

dabreegster (changes requested):

michaelkirk:
Just making sure  that new CI check is working, I swear! 😅😇 

https://github.com/a-b-street/ltn/actions/runs/14089745271/job/39463141898#step:6:140

Screenshot 2025-03-26 at 10 44 50

dabreegster (approved):
Thank you! New blueish color looks great

Issue #249, opened 2025-03-20
prepopulate first neighbourhood name with project name


michaelkirk:
#153  - we have a theory that most people will only have one neighbourhood per project, so we want to make it a little easier, especially for new users, to get to the neighbourhood editing part.

One thing that's annoying right now is you have to name your project, and then create your neighbourhood and then name your neighborhood. This makes sense in the "multiple neighbourhoods per project" universe, but when you just have one, they are almost inevitably the same thing.

Dustin's suggestion:

> I anticipate still a little user confusion about picking both a project and neighbourhood name, but we could maybe just prefill the first neighbourhood name with "LTN" or the project name.

Personally I like $`{project name} LTN`

And we could either:
1. pre-populate the prompt with this value, and have the user click confirm
2. or just skip the prompt and create the neighborhood with the automated name. The user can always go back and edit it.

Because neighbourhood names should be unique, for multiple neighbourhood projects, a couple UX alternatives
1. Don't engage the above auto-naming logic when there is already at least one neighbourhood in a project.
2. Double down on auto-naming and add a `${project_name} LTN #2`, knowing the user can edit it later.

Laying it all out, my best guess at a UI is:
1. pre-populate the prompt, have the user click confirm
2. For multi-neighbourhoods: double down on auto-naming and add a `${project_name} LTN #2`, but it's only in the prompt, so user can easily edit it before saving.

PR #271, opened 2025-03-25
Suggest default project and neighborhood names, don't allow picking blank.

### LINKED ISSUE #245: Project and neighborhood names should be trimmed of leading/trailing whitespace before saving.
Otherwise it's pretty easy to get confused:
Image

michaelkirk:
FIXES #245 and #249 

We now suggest project and neighbourhood names, which makes the creation process a bit faster and a bit less "daunting" if you don't know what these names are for.

**New flow:**

https://github.com/user-attachments/assets/0135d3d2-10f3-4b4c-bbd1-2646206e9ddf

We're also stricter about not accepting blank names:

https://github.com/user-attachments/assets/40563284-2460-4fa8-b674-54f1c18c121b

Followup work: the "names must be non-empty" check is only enforced in the UI at creation time. Notably missing are:
1. a migration for existing "invalid" data
2. requiring names for projects loaded from JSON.

I don't anticipate that either of these would be especially hard, but it seemed lower priority, given that we don't actually rely on the names being non-empty as far as I know.

### DIFF
--- a/web/src/AddNeighbourhoodMode.svelte
+++ b/web/src/AddNeighbourhoodMode.svelte
@@ -27,8 +27,10 @@
   import {
     appFocus,
     backend,
+    currentProjectID,
     map,
     mode,
+    projectStorage,
     returnToChooseProject,
     saveCurrentProject,
   } from "./stores";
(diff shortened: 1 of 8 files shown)


michaelkirk:
I've included #249, and this is ready for review.

dabreegster (approved):
All of this is brilliant! In quick demos, I always spam my keyboard randomly; this is so much better to not have to make up names all the time.

PR #272, opened 2025-03-25
CI runs ts type checking

michaelkirk:
My typescript editor (usually vscode) doesn't inspect files unless they're in an open tab, so I sometimes don't see type errors for a while. 

e.g I originally opened #270 with a [type error](https://github.com/a-b-street/ltn/compare/de7396eb277963afa16e2534d8ab94d0370da70c..67b748f312d337cc746d801162a59484b59bd159) (In this case, it wasn't a real error, just something too confusing for typescript).

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -51,6 +51,6 @@ jobs:
           cd web
           npm ci
           npm run wasm
-          # typechecking requires the wasm build
-          npm run check
           npm test
+          # typechecking requires the wasm build and for vite to see the build
+          npm run check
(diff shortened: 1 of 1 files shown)


michaelkirk:
I see that this check is failing... I might spend a little more time on this but TBH, it doesn't seem very high priority.

dabreegster:
The error is `Error: Cannot find module 'backend' or its corresponding type declarations. `

Which I also get when running locally, anytime you `npm run wasm` and then `npm check` before `npm run dev`. When you run dev, vite does something to actually see the generated WASM bindings. Possibly doing `check` after `test` could work, but agreed this is low-pri

dabreegster:
Hmm yeah, that trick works. OK to merge this now?

michaelkirk:
Brilliant, thanks!

PR #273, opened 2025-03-25
revisit prioritization

### LINKED ISSUE #235: Should be possible to "revisit" the prioritization data table shown on "pick boundaries" screen.
Currently you can only see this when building your neighbourhood:

Image

Presumably if you are using specific metrics to justify where your boundaries are, you should be able to re-visit and share those metrics with someone later.

We should add this... somewhere... I haven't thought about it very hard yet.

Maybe on the "Editing" page? It's not really about "editing", but that is kind of the only page where we are focused on a single neighbourhood. 

Or potentially on the "pick neighbourhood" page we could show the stats for all neighbourhoods. It's a lot of data, so maybe it'd need to be in a collapsed component or something.

We'll also have to consider this vis-a-vis #153

michaelkirk:
Fixes #235 

**before:**
Screenshot 2025-03-25 at 15 49 56

**after: cnt** reorganized content, and added a "stats" section

Screenshot 2025-03-25 at 15 54 58

**after: global** It's a bit sparse, with only "area" as metric, but maybe it's fine. 🤷 

Screenshot 2025-03-25 at 15 51 06


web/src/edit/ChangeFilterModal.svelte
dabreegster, line 1: Little confused by the rename, but no big deal

dabreegster (approved):
I love how clean the code change is, because the stats were already their own component. The new organization is looking great!

Global mode just having area is fine for now; I suspect I'll generalize a few of the Scottish metrics in the next few months. It's good placeholder space anyway.

web/src/edit/ChangeFilterModal.svelte
michaelkirk: It's confusing because it's a Modal (UI) about modal filters. 

I was confused at first why the filter control code was placed where it was in the code. I eventually realized that the component didn't render anything visible - except as a Modal, so its position in the code didn't really matter. With the rename I intended to capture this.

dabreegster: Ahhah, makes sense. Overloading strikes again.

PR #277, opened 2025-03-26
Fix occasionally missing hover state.

michaelkirk:
There seems to be a bug in svelte-maplibre (or maplibre?). Presumably the issue is related to quickly switching between hovered features.

To reproduce pan into and out of one of the CNT boundaries. About 10% of the time, the background will not be colored appropriately for the hover state. Interestingly, the popup will still appear.

https://github.com/user-attachments/assets/5d4d9f0f-6a60-4537-af16-f698343f2d43

I cannot reproduce the issue after removing "hover" handling from the linestring borders. We don't need it anyway.

There is a similar seeming bug in the autgen boundaries page where occasionally, an empty popup appears. It seems slightly less reproducible, but still fairly common to see. I haven't figured that one out yet, but stay vigilant 👀.

Here's what that (unresolved issue) looks like:

https://github.com/user-attachments/assets/45696952-ef2a-40f6-a2fa-eb4b6fb7c737

### DIFF
--- a/web/src/CntChooseArea.svelte
+++ b/web/src/CntChooseArea.svelte
@@ -112,6 +112,6 @@
       "line-width": 2.5,
     }}
     beforeId="Road labels"
-    manageHoverState
+    manageHoverState={false}
   />
 </GeoJSON>


dabreegster:
I've seen both problems...

(Just checking -- is this meant to be an issue?)

> About 10% of the time, the background will not be colored appropriately for the hover state. Interestingly, the popup will still appear.

I have not tested this extensively, but I think I've only seen it happen with large country-wide areas like this page. The polygons are quite complex and jagged. My vague theory is that there might be some issue with the triangulation, but it doesn't really make sense, because popups are fine. Other vague theory is that a symbol layer from the basemap (like labels) has focus and is stopping the hover state.

> an empty popup appears

I've seen this one in many places for a long time too. I _think_ it happens when you get a normal popup from hovering on the map, move the cursor quickly onto the popup, and lose focus on the map because the popup is in the way. But I'm not sure. It could also be related to sometimes wanting to hide a popup if for some reason Svelte reactivity is sluggish and a variable isn't initialized yet. I asked https://github.com/dimfeld/svelte-maplibre/discussions/242 a while ago and `canOpen` was the answer. It's in the latest svelte-maplibre, but since we're still stuck on svelte 4, I added it to my own branch https://github.com/dabreegster/svelte-maplibre/tree/svelte4_popups and am using in NPW in one place. Maybe that helps.

michaelkirk:
> Just checking -- is this meant to be an issue?

Sorry to be unclear - the issue highlighted in the first video is resolved by this PR.

It seems to affect not only "large LAD geometries" but will occur (intermittently) with any polygon layer which also has a corresponding border drawn as a separate **interactive** LineLayer.  I actually first noticed this behavior when introducing the border for the generated boundaries in the "add neighbourhood" mode. I fixed it there before merging, and then some days later recognized the behavior on the CNT front page. 


The bug in the second video I don't have a solution for. I'll create a followup issue to track that.

Issue #278, opened 2025-03-27
Bulk edit main road classification


dabreegster:
Apply one-way and main road changes to one road segment at a time can be painful. Some ideas for bulk selecting:

1) Group by contiguous road names. Maybe we produce a different GJ linestring dataset with these merged cases, and have the regular hover behavior, or maybe we keep the current per-road GJ data and just have a maplibre filter to show all matching road names

2) Route snapping tool, like the one used in route mode

3) A freehand paint tool, like "draw filters along line"

PR #283, opened 2025-03-27
tool layout

### LINKED ISSUE #275: Reorganize the "editing tools"
We talked about this previously, but I'm not sure it made it into an issue.

I think we could do some small things to make this toolbar clearer:

---
Image
---

- [ ] Move "draw along a long" and "change type" to be sub-options of "modal filters"
- [ ] With that much space, we could consider icon+text buttons

Also, those hover tooltips are partially off screen, but I'm not sure how easy that is to address.

michaelkirk:
Fixes #275 

I acknowledge that this is out of priority order, but I got a real itch for it.
As a fig leaf, it's somewhat related to #278, since I'd like to re-use the "paint" paradigm for my first attempt. I'm hoping since the line-drawing logic already exists, and users will see this paradigm in multiple places, it will (hopefully) be high benefit/cost.

## Screenshots

Screenshot 2025-03-27 at 16 24 10
Screenshot 2025-03-27 at 16 24 16
Screenshot 2025-03-27 at 16 24 20
Screenshot 2025-03-27 at 16 24 27

## Flow 

https://github.com/user-attachments/assets/cce6a305-71a3-4e55-91b1-9deaf9cdcf2e

(The tail end of that video shows what might be a bug: should the "reclassify main road" add to the undo stack? I think it should, but I forgot to add it when merging #218 )


michaelkirk:
Note: I tweaked some of the button colors in the last commit since the screenshots.

web/src/edit/NeighbourhoodMode.svelte
```diff
@@ -332,126 +330,157 @@
...
-        on:click={() => (action = { kind: "oneway" })}
-        disabled={action.kind == "oneway"}
-        class:active={action.kind == "oneway"}
-        class:outline={action.kind != "oneway"}
-        data-tooltip="Toggle one-way (hotkey 3)"
-      >
-        <!-- 
+    <div
+      class="tool-palette"
+      style="height: 54px; display: flex; justify-content: space-between; gap: 32px;"
```
dabreegster, line 335: On my end, the fixed height of the panel jumps around for some of the tools. The screenshot region below is the same, but you can see "Map style" jumps:
image
vs
image

This could be resolution or something else. Happy to look into this later.

web/src/edit/NeighbourhoodMode.svelte
```diff
@@ -332,126 +330,157 @@
...
-        data-tooltip="Ctrl+Y"
-      >
-        {#if redoLength == 0}
-          Redo
-        {:else}
-          Redo ({redoLength})
-        {/if}
-      </button>
+    <div
+      style="min-height: 200px; padding-bottom: 16px; border-bottom: solid var(--pico-muted-border-color) 1px;"
```
dabreegster, line 425: Oops, move that comment above down here. Maybe the min height needs to be a little higher. I'm guessing you used a browser inspector to measure the height of all cases, then picked the max for this value?

web/src/edit/NeighbourhoodMode.svelte
```diff
@@ -332,126 +330,157 @@
...
+            Change modal filter type
+          </button>
+
+          <button
+            on:click={action.freehand
+              ? () => (action = { kind: "filter", freehand: false })
+              : () => (action = { kind: "filter", freehand: true })}
+            class:active={action.freehand}
+            class:outline={!action.freehand}
+            data-tooltip="Add many modal filters along a line (hotkey 2)"
```
dabreegster, line 446:
```suggestion
            data-tooltip="Add many modal filters along a line"
```

The hotkey is gone now (which is a good choice)

dabreegster (approved):
This looks fantastic, thanks! Totally OK to work on things in whatever order; you will never  ever see me complain about UI polish or improved structuring. :smiley: 

> should the "reclassify main road" add to the undo stack? 

Ooh good catch -- yes, and it should be pretty easy.

dabreegster:
I snuck in 64a9b52549385ad839dad328c01e8771156a15dd after realizing the tooltip edit I made here was incomplete.

I had a quick look at undoing main roads. It's odd, the change goes through. If you go back to pick neighbourhood, then back, it shows up. But not immediately. It's an edit to the entire neighbourhood, but `mutationCounter` increases the same way on the frontend. I'll look in more later.

web/src/edit/NeighbourhoodMode.svelte
michaelkirk: It's the same on my machine. I actually didn't intend for there to be *no* difference in height, just to reserve enough space so the area was "recognizable" as the tab content area.

If we wanted no movement whatsover, we'd need to find our longest content at our narrowest width. Do you want me to make that change?

michaelkirk:
Thanks for the review and for the fixups!

Issue #287, opened 2025-03-30
Road classification hidden under arrows


XioNoX:
I haven't looked at LTN in a couple weeks, it's impressive to see the progressions !

Follow up from #218 

Small UI issue, when trying to re-classify a road into a secondary road, if some parts of it are small they get hidden under the "Area in/out" arrows, so it becomes a guessing game to know where to click.

A suggestion would be to add enough transparency to those arrows on mouse hover so the road itself becomes visible.

![Image](https://github.com/user-attachments/assets/4667b7d4-8cb3-43b9-b24d-00155089f55a)

dabreegster:
Thanks for reporting! This is a duplicate of #279. Michael, if you share a draft branch, I can look into the SDF icon idea too

Issue #279, opened 2025-03-27
"border arrows" are obtrusive (especially with interior main roads)


michaelkirk:
Followup to #218 

<img width="745" alt="Image" src="https://github.com/user-attachments/assets/2234600a-23ca-4dc2-9979-f706b2fe8c53" />

I have a half-baked approach to addressing this that switches to representing border arrows using a symbol layer, rather than a feature layer. This is nice because the icons will stay relatively the same screen size as you zoom in. Meaning that in the above example they would be much smaller.

TBH, I'm struggling a bit with the sdf icon + border (icon-halo) dance, but I have something that seems to work, if not a bit brittle.

dabreegster:
Have you figured out how to hide or scale down symbol size at low zoom? When you're far away, having lots of the symbols clustered next to each other gets pretty busy. I'm curious to see how this approach looks!

PR #288, opened 2025-03-30
Temporarily fade the border arrows to mitigate #279.

dabreegster:
@XioNoX just proposed slightly fading the border arrows to be able to see underneath them in #287. I tried this, and think it still looks legible but now lets you see grey / not-grey roads underneath:
image
Before the full approach in #279, how about we do this as a workaround. I'll ASAP merge this to unblock users having this problem, and we can adjust easily later.


XioNoX:
Wow, awesome, thanks! Especially for a Sunday morning!

michaelkirk:
Seems like a good quick fix!

PR #291, opened 2025-03-31
FIX: adding diagonal filter also adds modal filter

### LINKED ISSUE #290: Bug: Adding diagonal filter also adds modal filter
I'm not sure when this regressed, but I can take a look soon.

https://github.com/user-attachments/assets/ebe53296-1d30-45cb-8d09-d7882f8dc2e9

michaelkirk:
FIXES #290 

This is a partial revert of https://github.com/a-b-street/ltn/commit/49629fec0a0d543a69929b9cd64e5ba5115eb464#diff-eb72a5e7f5a0b04e27ba1e11c412351855e54f41c1ca662fb0556dc652d818a5R100

Which introduced manually marking/unmarking a road as a "main road".

@dabreegster - do you recall why adding this option was necessary. In my testing I'm still able to mark/unmark roads without this option. If we need it, I'll have to come up with another way to resolve the issue.

### DIFF
--- a/web/src/layers/NeighbourhoodRoadLayer.svelte
+++ b/web/src/layers/NeighbourhoodRoadLayer.svelte
@@ -102,7 +102,7 @@
 />
 
 <LineLayer
-  {...layerId("interior-roads", false)}
+  {...layerId("interior-roads")}
   filter={["==", ["get", "kind"], "interior_road"]}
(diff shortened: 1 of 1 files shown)


michaelkirk:
Alright, great, let's roll with it.

PR #292, opened 2025-04-01
Rewrite project credits

### LINKED ISSUE #79: Update project credits
There'll be more to add soon, but in the meantime, @xioNoX and @rouelibre1, you've both been incredible at testing / feedback / ideas and I'd love to list you in credits. Is that OK, and is just linking to your GH profile the best way?

dabreegster:
FIXES #79

### DIFF
--- a/web/src/About.svelte
+++ b/web/src/About.svelte
@@ -5,46 +5,88 @@
 
 <Modal bind:show={$showAbout}>
   <h1>The low-traffic neighbourhood (LTN) tool, v2</h1>
+
   <p>
-    This is an <b>experimental</b>
-    version of the
-    <a
-      href="https://a-b-street.github.io/docs/software/ltn/index.html"
-      target="_blank"
+    This is the second version of the A/B Street LTN tool, <a
+      href="https://github.com/a-b-street/ltn"
+      target="_blank">an open source project</a
+    >. This version, called the Connected Neighbourhoods Tool, has been funded
+    by
+    <a href="https://www.transport.gov.scot/" target="_blank"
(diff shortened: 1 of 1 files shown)


web/src/About.svelte
```diff
@@ -5,46 +5,88 @@
...
     >
-      A/B Street LTN tool
-    </a>
-    . Most parts of it do not work yet, and you should probably use the other version
-    of the tool instead.
+    and designed by
+    <a href="https://www.sustrans.org.uk" target="_blank">Sustrans Scotland</a>.
+    The team includes:
   </p>
-  <p>To use this tool, you need to:</p>
```
dabreegster, line 20: This screen isn't the right place for instructions, and the new design of the app makes the flow so much more clear anyway

web/src/About.svelte
```diff
@@ -5,46 +5,88 @@
...
+            >Congying (Cici) Hu</a
+          >, mobility planner
+        </li>
+        <li>
+          <a
+            href="https://www.linkedin.com/in/angus-calder-a69543206"
+            target="_blank">Angus Calder</a
+          >, senior mobility planner
+        </li>
+        <li>Michael Naysmith, senior mobility planner (PM)</li>
```
michaelkirk, line 47: A minor thing, but is it well understood what PM means in this context?

michaelkirk (approved):
LGTM!

web/src/About.svelte
dabreegster: Good point; Michael put this in the google doc directly, but expanding the acronym to be more clear

PR #294, opened 2025-04-02
Refactor to prepare for more accurate routing

dabreegster:
Preparation for #243, internally changing the API to route from more accurate positions along a road and plumbing the necessary router input everywhere. There shouldn't be any behavioral diff yet, but the API change is big enough to warrant a first PR to make the second easier to understand.

### DIFF
--- a/backend/benches/router.rs
+++ b/backend/benches/router.rs
@@ -30,7 +30,8 @@ fn benchmark_route(c: &mut Criterion) {
     ] {
         let map = neighbourhood.map_model().unwrap();
         let main_road_penalty = 1.0;
-        let router = Router::new(&map.router_input_before(), main_road_penalty);
+        let router_input = map.router_input_before();
+        let router = Router::new(&router_input, main_road_penalty);
 
         let route_requests = synthetic_od_requests(&map);
         c.benchmark_group(neighbourhood.savefile_name)
@@ -39,7 +40,7 @@ fn benchmark_route(c: &mut Criterion) {
                 b.iter(|| {
                     let mut num_found = 0;
                     for (start, end, _) in &route_requests {
-                        if let Some(_found) = router.route_from_roads(*start, *end) {
+                        if let Some(_found) = router.route_from_roads(&router_input, *start, *end) {
                             num_found += 1;
                         }
                     }
--- a/backend/src/route.rs
+++ b/backend/src/route.rs
@@ -2,11 +2,14 @@ use std::cell::RefCell;
 use std::collections::HashMap;
 
 use fast_paths::{FastGraph, InputGraph, PathCalculator};
-use geo::{Coord, Euclidean, Length, LineString};
+use geo::{Coord, Euclidean, Length, LineLocatePoint, LineString};
+use rstar::{primitives::GeomWithData, RTree};
 use utils::NodeMap;
 
 use crate::map_model::{DiagonalFilter, Direction};
-use crate::{Intersection, IntersectionID, MapModel, ModalFilter, Road, RoadID, TravelFlow};
+use crate::{
+    Intersection, IntersectionID, MapModel, ModalFilter, Position, Road, RoadID, TravelFlow,
+};
 
 // For vehicles only
 pub struct Router {
@@ -16,7 +19,7 @@ pub struct Router {
     pub main_road_penalty: f64,
 }
 
-#[derive(Debug, Clone, PartialEq, Eq)]
+#[derive(Debug, Clone)]
 pub struct Route {
     pub steps: Vec<(RoadID, Direction)>,
 }
@@ -27,6 +30,7 @@ pub struct Route {
 /// specify a subset of roads, so we can build a routing graph focused within a single neighborhood.
 pub trait RouterInput {
     fn roads_iter(&self) -> impl Iterator<Item = &Road>;
+    fn closest_road(&self) -> &RTree<GeomWithData<LineString, RoadID>>;
 
     fn get_r(&self, r: RoadID) -> &Road;
     fn get_i(&self, i: IntersectionID) -> &Intersection;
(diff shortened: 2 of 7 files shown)


backend/benches/router.rs
```diff
@@ -39,7 +40,7 @@ fn benchmark_route(c: &mut Criterion) {
                 b.iter(|| {
                     let mut num_found = 0;
                     for (start, end, _) in &route_requests {
-                        if let Some(_found) = router.route_from_roads(*start, *end) {
+                        if let Some(_found) = router.route_from_roads(&router_input, *start, *end) {
```
dabreegster, line 43: The danger is accidentally using the wrong `RouterInput` input for a `Router`. I'm not sure of a reasonable way to express "when the map model changes, the router gets invalidated" in Rust lifetimes. Does this seem OK to you?

backend/src/route.rs
```diff
@@ -37,6 +41,20 @@ pub trait RouterInput {
...
     fn turn_restrictions(&self, i: IntersectionID) -> &Vec<(RoadID, RoadID)>;
+
+    fn snap_to_road(&self, pt: Coord) -> Position {
+        let r = self
+            .closest_road()
+            .nearest_neighbor(&pt.into())
+            .unwrap()
+            .data;
+        let road = self.get_r(r);
+        let percent_along = road.linestring.line_locate_point(&pt.into()).unwrap();
```
dabreegster, line 52: This extra step is slower. I'm getting noisy results from just rerunning the benchmarks on main, but here's what they say against this branch:

```
dundee/predict impact   time:   [471.75 ms 474.57 ms 478.63 ms]
                        change: [-25.834% -15.751% -6.6921%] (p = 0.01 < 0.05)
                        Performance has improved.
bristol_east/routing    time:   [7.1597 ms 7.7003 ms 8.3021 ms]
                        change: [+14.274% +22.807% +31.185%] (p = 0.00 < 0.05)
                        Performance has regressed.
strasbourg/routing      time:   [18.042 ms 18.261 ms 18.665 ms]
                        change: [-3.3327% +0.0369% +4.1819%] (p = 0.99 > 0.05)
                        No change in performance detected.
shortcuts in bristol_east
                        time:   [2.5329 ms 2.5668 ms 2.6034 ms]
                        change: [+4.9188% +6.3439% +7.9653%] (p = 0.00 < 0.05)
                        Performance has regressed.
shortcuts in bristol_west
                        time:   [11.330 ms 11.836 ms 12.371 ms]
                        change: [+15.331% +20.644% +26.417%] (p = 0.00 < 0.05)
                        Performance has regressed.

shortcuts in strasbourg time:   [3.6981 ms 3.7599 ms 3.8270 ms]
                        change: [+3.0100% +4.8524% +6.7360%] (p = 0.00 < 0.05)
                        Performance has regressed.
```

I have no idea why some benchmarks _improve_ or stay the same. I'd expect anything calling route to get a bit slower.

backend/src/route.rs
```diff
@@ -121,13 +139,45 @@ impl Router {
...
+            Position {
+                road: end,
+                percent_along: 0.5,
+            },
+        )
+    }
+
+    pub fn route_from_positions(
+        &self,
+        _router_input: &impl RouterInput,
```
dabreegster, line 177: This PR isn't yet adding the behavior of paying attention to the position along the road. But we will need RouterInput to do that -- when `starts` or `ends` is empty, it means there's a modal filter on the start or end road. We need to get the modal filter, to know its position on the road. We also need to look up roads and intersections to figure out which way we can search.

backend/src/shortcuts.rs
```diff
@@ -84,10 +87,11 @@ impl Shortcuts {
 
                         // How long is the shortest route through the original router, using this
                         // neighbourhood or not?
-                        let direct_length = match map
-                            .router_before
-                            .route_from_roads(*start_r, *end_r)
-                        {
+                        let direct_length = match map.router_before.route_from_roads(
```
dabreegster, line 90: This is subtly wrong, about to fix...

backend/benches/router.rs
michaelkirk: To make sure I'm understanding: 

The router graph's behavior is predicated on its input, so if its input changes, we need to rebuild the router. 

But that's an existing requirement, right? Or is there some new requirement related to that coming from #243?

dabreegster: Your understanding is correct -- it is an existing requirement. But now for #243, we need to sometimes check what roads have a modal filter, and get its precise position. The new danger is calling `map.router_before(map.router_input_after(), pos1, pos2)` accidentally.

michaelkirk: Ah, I see! Yeah, that's a little rough... but I see why you came to this conclusion.

I think it's reasonable enough and we should go forward with it. Maybe we can revisit simplifying the code once we have #243 in place.

backend/src/shortcuts.rs
```diff
@@ -84,10 +89,11 @@ impl Shortcuts {
...
```
michaelkirk, line 92: What's the significance of switching routers here?

michaelkirk (approved):
LGTM, thanks for making a smaller PR that was easy to follow.

backend/src/shortcuts.rs
dabreegster: I'm actually not sure. I almost missed this at first and plugged `router_input_after`.

The directness is just for sorting and showing the "most likely" shortcuts first. If a shortcut through a neighbourhood is 100m long and the route using the entire map is 10m, then it's not a likely shortcut. The intention is just to see if using the main roads changes the distance a bunch. So maybe it's fine to use the full `map.router_after`? I think the reason that isn't done right now is performance -- we only call `rebuild_router` when absolutely necessary. Doing it every time you make any edit to a neighbourhood and need to recalculate shortcuts gets expensive.

Issue #121, opened 2025-02-12
Branding / Design Grab bag


michaelkirk:
This issue was originally a placeholder for any CNT specific branding, but instead became a grab-bag of UI improvements, so updating the description accordingly:

---
- [x] narrower top bar 

<img width="2066" alt="Image" src="https://github.com/user-attachments/assets/b0ea22ca-0976-4617-847b-8968cbf0d8f8" />

--- 
- [x] Move basemap toggle onto map control

<img width="704" alt="Image" src="https://github.com/user-attachments/assets/00b5544e-9719-40be-864f-32087e93c25d" />

Maybe we can get rid of settings all together?

- [x] TBH I haven't used "developer mode" yet, so I'm not sure what makes sense. Maybe it could be a URL flag? https://github.com/a-b-street/ltn/pull/301/files

---

- [x] "Road thickness depends on shortcuts" only makes sense in the "edit" mode. (I think?) in #296

<img width="704" alt="Image" src="https://github.com/user-attachments/assets/adb4a8d0-a74b-48ef-8701-eacbce724861" />

I think we should move it to "draw roads"

---

- [ ] These buttons seem a bit out of place. I haven't really thought about their resolution.

<img width="400" alt="Image" src="https://github.com/user-attachments/assets/4f41a929-34f0-4a34-81ac-4bd9d3bc6792" />

The bottom 2 seem like "map navigation tools", and they are visible on every screen (except title mode), so maybe it'd make sense for them to be *on* the map, like the other map navigation tools.

<img width="260" alt="Image" src="https://github.com/user-attachments/assets/f4f12604-8a1c-4120-9bc1-f019dd962fca" />

michaelkirk:
Are there any specific things you'd like to see done here?

I know there are definitely some design things I'd like to address, some of them could probably be cribbed from npw.

Here's a laundry list:

narrower top bar 
<img width="2066" alt="Image" src="https://github.com/user-attachments/assets/b0ea22ca-0976-4617-847b-8968cbf0d8f8" />

Move basemap toggle onto map control

<img width="704" alt="Image" src="https://github.com/user-attachments/assets/00b5544e-9719-40be-864f-32087e93c25d" />

Maybe we can get rid of settings all together?

"Road thickness depends on shortcuts" only makes sense in the "edit" mode. (I think?)

<img width="704" alt="Image" src="https://github.com/user-attachments/assets/adb4a8d0-a74b-48ef-8701-eacbce724861" />

[...]

PR #296, opened 2025-04-02
move road thickness "setting" to neighourhood -> "map style"

michaelkirk:
Part of #121 

This setting only affects the map in NeighbourhoodMode anyway. It fits nicely along side options like "Animate shortcuts" and "Show entires into cells".

I also applied a "footnote" style to the Cells and Shortcut "definitions" in the Map style copy.

**before:** Road thickness depends on shortcuts was in the "global settings" menu in the title bar.

Screenshot 2025-04-02 at 16 14 52

**before:** Map style within neighborhoods:
 
Screenshot 2025-04-02 at 16 15 01

**After:** "Road thickness..." moved into map style

Screenshot 2025-04-02 at 16 14 17

Also, put the "road style" control on a single line, which obviates the need for a border.

I also restyled the definitions as "footnotes", since they were a little awkward inline, and could apply to multiple settings. And tweaked the cells copy to mention "colored". WDYT?

### DIFF
--- a/web/src/edit/NeighbourhoodMode.svelte
+++ b/web/src/edit/NeighbourhoodMode.svelte
@@ -48,6 +48,7 @@
     returnToChooseProject,
     roadStyle,
     saveCurrentProject,
+    thickRoadsForShortcuts,
   } from "../stores";
   import type {
     NeighbourhoodBoundaryFeature,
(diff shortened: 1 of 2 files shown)


web/src/edit/NeighbourhoodMode.svelte
```diff
@@ -561,43 +562,53 @@
...
       interior.
     </p>
-
-    <label>
-      <input type="checkbox" bind:checked={$drawBorderEntries} />
-      Show arrows into cells
-    </label>
     <p>
-      <i>Cells</i> show the area reachable without travelling along a main road.
+      <sup>[2]</sup>
```
dabreegster, line 607: The footnotes take lots of space. I'm (slowly) investigating something like https://atomiks.github.io/tippyjs/ for NPW; tooltips might be a good approach here. The footnote notation is nice and recognizable for "there's more context here, maybe hover over?"

dabreegster (approved):

web/src/edit/NeighbourhoodMode.svelte
michaelkirk: Good idea! Added in #302

PR #298, opened 2025-04-03
Move basemap layer control from settings into contextual layers menu.

michaelkirk:
Based on #296, so review that first.

Part of #121 

**before:** basemap was in the global settings in the title bar

Screenshot 2025-04-02 at 17 27 06

**after:** basemap picker is in the layers control

**cnt:**
Screenshot 2025-04-02 at 17 23 05

**global:**
Screenshot 2025-04-02 at 17 23 36

I think that looks pretty good, but I'm not 100% sure we should proceed. One thing that didn't occur to me, which is nice about the current settings control, is that it's visible on the title page, before you pick a project.

I've maintained that, by making the contextual layers available on the title page, but maybe it's weird that the layers menu has different content depending on whether you're looking at a project or on the title page. WDYT?

Screenshot 2025-04-02 at 17 22 54

I'm slightly biased towards merging this, but wouldn't be mad if we didn't.


dabreegster (approved):
This makes sense for placement!

I think it would've been fine to leave the user stuck with the last chosen basemap on the title screen, but the consistency of changing it anywhere is nice -- thanks.

As you've mentioned in the issue, this just leaves the nebulous "developer mode" in settings. Maybe that could just be a URL query param?

PR #301, opened 2025-04-03
Replace $devMode setting with ?dev=true url param.

michaelkirk:
It was the last item in Settings, so let's get rid of the Settings modal all together.

Screenshot 2025-04-03 at 06 31 06


web/src/App.svelte
```diff
@@ -72,6 +72,17 @@
         console.log("Using local cache, not od2net.org");
       }
     } catch (err) {}
+
+    let devParam = new URLSearchParams(window.location.search).get("dev");
+    if (devParam === "true" || devParam === "") {
```
michaelkirk, line 77: http://localhost:5173/ltn/?dev=true or http://localhost:5173/ltn/?dev

dabreegster (approved):
Thank you for driving this simplification of the UI!

PR #308, opened 2025-04-04
Move map controls to map

### LINKED ISSUE #121: Branding / Design Grab bag

michaelkirk:
FIXES #121 (last item in the grab bag)

**before:** these controls were on every screen in the left panel
Screenshot 2025-04-03 at 17 07 58

**after:** Now they're on the map panel, along side our other map controls

Screenshot 2025-04-03 at 17 06 26

**flow:**

https://github.com/user-attachments/assets/e4b8376e-8ad8-450f-a1ad-7680bc1b38f1

Changing the "eye" icon when the configuration is presented is somewhat silly, but I think it helps communicate that the button has two states. Click it again to close it and return to the default state.

### DIFF
--- a/web/src/App.svelte
+++ b/web/src/App.svelte
@@ -7,10 +7,14 @@
   import ContextualLayers from "./context/ContextualLayers.svelte";
   import "@picocss/pico/css/pico.conditional.jade.min.css";
   import initLtn from "backend";
+  import { House } from "lucide-svelte";
   import type { LngLatBoundsLike, Map, StyleSpecification } from "maplibre-gl";
   import { init as initRouteSnapper } from "route-snapper-ts";
   import { onMount } from "svelte";
   import {
+    Control,
+    ControlButton,
+    ControlGroup,
     FillLayer,
     GeoJSON,
     MapLibre,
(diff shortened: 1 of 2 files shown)


dabreegster (approved):
Thank you, this works great!

CC @mvl22, who's working on NPW designs and was exploring options for streetview controls for a local access mode. Maybe the eye open/closed design is worth using there too. You can play with this tool at https://a-b-street.github.io/ltn/cnt.html by the way.

PR #309, opened 2025-04-04
Layer control: small design improvements

### LINKED ISSUE #276: Context button header should stay visible - content should scroll behind it.
As our context bar has gotten longer, this is becoming more of an  issue:

Image

I think that "layers" part should always remain visible.

Also, and maybe this should be a separate issue, sometimes it's unclear when you expand the bottom items that new content is available. We should adjust the scroll position.

https://github.com/user-attachments/assets/3ac45b96-45e2-4a9e-9670-dc29ffa2bdbe

michaelkirk:
FIXES #276 

**Before:** Scrolling down would scroll the header out of view, so it was harder to minimize the layer controls

Screenshot 2025-04-03 at 18 29 00

**After:** Content scrolls behind the header, so you can always minimize the layers
Screenshot 2025-04-03 at 18 35 51

I also changed the color scheme to be more neutral - the previous colors were based on the "pico secondary" button styles which feel very heavy to me. 

Screenshot 2025-04-03 at 18 35 46

I don't feel super strongly about that part of the change though. I put it into a separate commit so you could easily revert it if you prefer something closer to the old colors.

### DIFF
--- a/web/src/common/ContextLayerButton.svelte
+++ b/web/src/common/ContextLayerButton.svelte
@@ -6,14 +6,14 @@
 </script>
 
 <button
+  class="context-control"
   style="display: flex; align-items: center; justify-content: leading; gap: 8px; text-align: left;"
-  class="secondary"
   on:click={() => (show = !show)}
 >
   <input style="aspect-ratio: 1.0" type="checkbox" bind:checked={show} />
   {label}
   {#if $$slots.help}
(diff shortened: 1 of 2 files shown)


dabreegster (approved):
The changes all look great! The previous secondary coloring was indeed strong; I half thought it was leftover from pico's dark mode somehow

PR #312, opened 2025-04-04
refactor: dedupe/consolidate linking to mode

michaelkirk:
In preparation for #280 

This is *mostly* a no-op refactor, with a couple modest changes. I'll call out inline.

PR #313, opened 2025-04-04
better breadcrumbs

### LINKED ISSUE #280: Breadcrumb labels could be clearer
The "Pick Neighborhood" mode has evolved to also include seeing your overall project. I think the breadcrumb labeling should reflect this. 

Currently the breadcrumbs look like: `Choose Project > Pick Neighbourhood > Edit`

Image

Simplest would be to change "Pick Neighborhood" to "Project Overview"

If I had my druthers though, we'd do something like  `Choose Project > Dundee > Dundee LTN #2`

For pathologically long project names, like "Some Project with a long name", we'd have to truncate to something like `Choose Project > Some Project wi... > Dundee LTN #2`

michaelkirk:
FIXES #280 

Based on #312, so review that first.

https://github.com/user-attachments/assets/e0ebef46-e055-4a4d-9a79-9ab104c3f8bc

(🤦 Note how I tried to click "Choose Area" instead of "Create neighborhood" at 00:12s. I should copy the "active tool" style from the edit tools, I'll do that right now, should be quick).

Here's how the breadcrumbs look with a pathologically long name on a very narrow screen:

If it's an "inner" breadcrumb, we truncate it:

Screenshot 2025-04-04 at 15 49 28

If it's the final breadcrumb (the current page), we unfurl the full name in all its glory - so the user can see the full name.

Screenshot 2025-04-04 at 15 49 22

Not great... but there's only so much we can do!

Note: The max-size is defined in units of `vw`, so we'll truncate less on wider screens.

---

Also: With the project and neighbourhood names in the breadcrumb, their appearance elsewhere was feeling redundant, so I removed them.

Here's what it looked like after adding the names to the breadcrumbs, but before removing the redundancy:

**before**:

Screenshot 2025-04-04 at 15 55 04

**after**:

Screenshot 2025-04-04 at 15 56 51

### DIFF
--- a/backend/src/lib.rs
+++ b/backend/src/lib.rs
@@ -203,8 +203,8 @@ impl LTN {
     }
 
     /// `input`: GeoJson Feature w/ Polygon Geometry
-    #[wasm_bindgen(js_name = setNeighbourhoodBoundary)]
-    pub fn set_neighbourhood_boundary(
+    #[wasm_bindgen(js_name = setCurrentNeighbourhoodBoundary)]
+    pub fn set_current_neighbourhood_boundary(
         &mut self,
         name: String,
         neighborhood_feature: JsValue,
@@ -215,7 +215,9 @@ impl LTN {
             NeighbourhoodDefinition::from_feature(feature, &self.map).map_err(err_to_js)?;
         let boundary =
             NeighbourhoodBoundary::new(neighbourhood_definition, self.map.context_data.as_ref());
-        self.map.boundaries.insert(name, boundary);
+        self.map.boundaries.insert(name, boundary.clone());
+
+        self.neighbourhood = Some(Neighbourhood::new(&self.map, boundary).map_err(err_to_js)?);
         Ok(())
     }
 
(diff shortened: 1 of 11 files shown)


web/src/common/ProjectStorage.ts
```diff
@@ -234,9 +234,9 @@ export class ProjectStorage {
   }
 
   nextAvailableNeighbourhoodName(projectID: ProjectID): string {
-    let project = this.project(projectID);
-    let basename = `${project.project_name} LTN`;
+    let basename = "My neighbourhood";
```
michaelkirk, line 237: Is "My" too corny?

As I said in the description, with the project name in the breadcrumbs, also including it here felt redundant.

But just calling it "LTN" felt a bit sparse. I also don't know if that abbreviation is meaningful enough to everyone who might use the app.

Anyway, open to suggestions here!

backend/src/lib.rs
```diff
@@ -244,6 +246,13 @@ impl LTN {
         self.neighbourhood =
             Some(Neighbourhood::new(&self.map, boundary.clone()).map_err(err_to_js)?);
 
+        // We can delete this assert if it's a valid code path, but I think it's not.
```
dabreegster, line 249: Oh good spot, yeah. I forgot about this. I think you're right; this seems impossible now

web/src/common/ProjectStorage.ts
dabreegster: This seems fine to me. We can get feedback this next week on it.

web/src/common/navbar.ts
dabreegster, line 1: Ah, ignore my suggestion to use a constant string->string map from the other PR then

dabreegster (approved):

PR #318, opened 2025-04-08
Simplify generated waypoints from "picked" neighbourhoods.

### LINKED ISSUE #201: Simplify the generated waypoints for autogenerated boundaries
Currently we just make a waypoint for every vertex, which *works* but is not very friendly.

Image

michaelkirk:
FIXES #201 

You can create your neighbourhood "by hand" or by picking from a set of generated "severance" blocks.

Typically starting with the generated severance blocks is easier, but it's common that you also want to tweak the boundary a little bit to include/exclude some roads on the edge.

Adjust boundaries means manipulating the placement of Waypoints. For hand-drawn neighbourhoods, this is easy - we store the Waypoints placed by the user when they were drawing the neighbourhood initially.

For generated boundaries this is trickier. Previously, we simply generated a waypoint for every single vertex in the polygon. Because these shapes could be a bit complex, we'd easily end up with hundreds of waypoints. In practice, editing one of these neighbourhoods was very tedious.

The change here, is to apply a line simplification to the border, greatly reducing the number of waypoints.

Simplification, by it's nature, removes some fidelity of the polygon's exterior, so it's possible that we'll add/lose some roads on the edge of a boundary. My own testing shows this to be minimal and usually ignorable.

When it *does* introduce a non-ignorable problem, the user will need to go back and further refine the boundary.

We could further reduce the possibility of error by decreasing the `epsilon` value for the simplification algorithm, but that also means including more points.

In my own testing, I feel like the benefits of having a simpler polygon to start with outweighs the downsides of any jitter on the edge, and the current value is about right.

Before on the left (350+ points) after on the right (30ish points)

Screenshot 2025-04-08 at 10 53 32

### DIFF
--- a/backend/src/lib.rs
+++ b/backend/src/lib.rs
@@ -15,6 +15,7 @@ pub use self::route::Router;
 pub use self::shortcuts::Shortcuts;
 use crate::geo_helpers::make_polygon_valid;
 use crate::map_model::{Command, ProjectDetails};
+use crate::neighbourhood::WayPoint;
 use geo::{Coord, LineString, Polygon};
 use geojson::{Feature, FeatureCollection, GeoJson, Geometry};
 use serde::Deserialize;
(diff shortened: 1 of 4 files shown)


michaelkirk:
Here are some examples I've encountered while testing where adjusting changes the road network:

image

insignificant spurs added - I don't think anybody would notice or care:

Screenshot 2025-04-08 at 11 09 26

I haven't yet found an example where things are changed significantly, though I'm sure it's possible if you find just the right layout - e.g. maybe you can inadvertently shave off a critical Main road section. 

The consequences of this kind of clipping should be pretty obvious to the user though, and since they were just in the "adjust boundary" tool, they should know how to fix it.

backend/src/lib.rs
```diff
@@ -202,6 +203,19 @@ impl LTN {
         Ok(serde_json::to_string(&merged_boundary.to_feature(&self.map)).map_err(err_to_js)?)
     }
 
+    #[wasm_bindgen(js_name = extractWaypointsFromRing)]
+    pub fn extract_waypoints_from_polygon(&self, js_polygon: JsValue) -> Result<String, JsValue> {
+        let geojson_geometry: Geometry = serde_wasm_bindgen::from_value(js_polygon)?;
+
+        let Ok(mut ring) = geo::LineString::try_from(geojson_geometry) else {
+            return Err("invalid Polygon GeoJSON".into());
```
dabreegster, line 211:
```suggestion
            return Err("invalid LineString GeoJSON".into());
```

dabreegster (approved):
This is a simple and effective workaround -- nice. I think we could do better someday by generating minimal snapped waypoints, but that larger refactor is not worth it yet. Thanks!

michaelkirk:
Yeah, snapped waypoints would be nicer - it's usually what I want anyway.

dabreegster:
Just noting an example in Aberdeen where the simplification seems to pick up a parallel road just outside the boundary, and count it as a disconnected cell:
image
image
By https://www.openstreetmap.org/#map=19/57.150896/-2.172039

Not a big deal, easy for the user to fix

PR #319, opened 2025-04-08
Link to "predict impact" mode from "neighbourhood" mode.

### LINKED ISSUE #281: "Predict Impact" hard to find
From a data model perspective, "impact" is measured at the level of the project, not the neighborhood.

So it makes some sense that "predict impact" is accessible from the project overview (aka "Pick neighbourhood" mode).

In common usage, users will want to switch between "editing" and seeing the impact of that editing.

One proposal was to redundantly link to the "Project Impact" from the "editing" mode.

Also #280 should help a bit.

michaelkirk:
FIXES #281 

One shortcut I took is opting out of tracking the "true" back state for
the DebugDemandMode, accessible via PredictImpactDetailMode. We just
always go back to pick-neighbouhood. I don't think people will get very
confused by this in practice.

If we want to continue to simulate a stack properly so that hitting "back" 3 times
returns you to the same page from DebugDemandMode, we'd need to track
`prevPrevPrevMode` which seems a bit much.

At some point it might make sense to track an explicit stack of previous
`modes`, rather than trying to manually track the "prevMode" and
"prevPrevMode".

### DIFF
--- a/backend/src/lib.rs
+++ b/backend/src/lib.rs
@@ -246,13 +246,6 @@ impl LTN {
         self.neighbourhood =
             Some(Neighbourhood::new(&self.map, boundary.clone()).map_err(err_to_js)?);
 
-        // We can delete this assert if it's a valid code path, but I think it's not.
-        // If we haven't triggered it after a while, we can delete the editing_same logic.
-        debug_assert!(
-            !editing_same,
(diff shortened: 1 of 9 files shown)


dabreegster:
The tradeoffs made here LGTM. I don't think we will/should ever have such a deeply nested state and need to use a real stack. We don't want to wind up with widgetry, after all ;)

PR #321, opened 2025-04-08
Recreation POIs

### LINKED ISSUE #233: Add "Recreation" POI category
This should include (at least) pubs and cafes.

michaelkirk:
FIXES #233 

We should have sustrans take a sanity pass at our categories.

Screenshot 2025-04-08 at 16 44 17

Zooming in you can see how the recreation POI's often correspond with the map data POIs
Screenshot 2025-04-08 at 16 44 31

I was tempted to add restaurants and fast_food, as they are likely destinations, but I couldn't justify filing taco bell under recreation.

### DIFF
--- a/data_prep/scotland/get_reference_layers.sh
+++ b/data_prep/scotland/get_reference_layers.sh
@@ -44,6 +44,28 @@ function cbd {
   wget https://nptscot.blob.core.windows.net/pmtiles/cbd_layer_2024-12-01.pmtiles -O $OUT/cbd.pmtiles
 }
 
+function recreation {
+  get_scotland_osm
+
+  # This list of tags is somewhat arbitary.
+  # I went down the amenity,leisure,tourism tags and picked the most popular ones that arguably fall under recreation.
+  # Note that we're currently only considering node geometries. This will exclude items which are defined on an area
+  # or linear geometry.
+  osmium tags-filter tmp/scotland-latest.osm.pbf \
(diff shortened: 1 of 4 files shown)


data_prep/scotland/get_reference_layers.sh
dabreegster, line 1: I was briefly thinking we could scrape these while we parse the OSM anyway, but that's totally at odds with mixing into the pre-built prioritization data. I like this more explicit approach.

dabreegster (approved):
LGTM! New layer is uploaded, and so are the regenerated prioritization files.

> I was tempted to add restaurants and fast_food, as they are likely destinations, but I couldn't justify filing taco bell under recreation.

Ah, the subjective choices we make helping transport modellers ;)

data_prep/scotland/get_reference_layers.sh
```diff
@@ -44,6 +44,28 @@ function cbd {
...
 }
 
+function recreation {
+  get_scotland_osm
+
+  # This list of tags is somewhat arbitary.
+  # I went down the amenity,leisure,tourism tags and picked the most popular ones that arguably fall under recreation.
+  # Note that we're currently only considering node geometries. This will exclude items which are defined on an area
+  # or linear geometry.
+  osmium tags-filter tmp/scotland-latest.osm.pbf \
```
dabreegster, line 54: @congying-hu, here are the OpenStreetMap categories for recreation POIs for both the layer and land use density metric. The POI layer at https://a-b-street.github.io/ltn/cnt.html should have a new dot showing these. If it looks like there are any missing, we can update this list.

PR #323, opened 2025-04-09
improve predict impact perf

michaelkirk:
Some progress on #129 - I'd say this covers the low and medium hanging fruit anyway.

There's a few fundamental changes here, which I'll call out in line.

I'm going to leave these commits unsquashed for now in case you were curious or, if some of the changes are controversial, to make them more easily removable.

### DIFF
--- a/backend/benches/impact.rs
+++ b/backend/benches/impact.rs
@@ -1,7 +1,25 @@
 use backend::test_fixtures::NeighbourhoodFixture;
 use criterion::{criterion_group, criterion_main, Criterion};
 
-fn benchmark_impact(c: &mut Criterion) {
+fn benchmark_impact_initial_build(c: &mut Criterion) {
+    for fixture in [NeighbourhoodFixture::DUNDEE] {
+        let mut map = fixture.map_model().unwrap();
+        let fast_sample = false;
+
+        c.benchmark_group(fixture.savefile_name)
+            .sample_size(10)
+            .bench_function("predict impact - initial build", |b| {
+                b.iter(|| {
+                    map.rebuild_router(1.0);
+                    let mut impact = map.impact.take().unwrap();
+                    impact.recalculate(&map, fast_sample);
+                    map.impact = Some(impact);
+                })
+            });
+    }
+}
+
+fn benchmark_impact_rebuild(c: &mut Criterion) {
     for fixture in [NeighbourhoodFixture::DUNDEE] {
         let mut map = fixture.map_model().unwrap();
         let fast_sample = false;
@@ -16,7 +34,7 @@ fn benchmark_impact(c: &mut Criterion) {
 
         c.benchmark_group(fixture.savefile_name)
             .sample_size(10)
-            .bench_function("predict impact", |b| {
+            .bench_function("predict impact - rebuild", |b| {
                 b.iter(|| {
                     let mut impact = map.impact.take().unwrap();
                     impact.invalidate_after_edits();
@@ -27,5 +45,9 @@ fn benchmark_impact(c: &mut Criterion) {
     }
 }
 
-criterion_group!(benches, benchmark_impact);
+criterion_group!(
+    benches,
+    benchmark_impact_rebuild,
+    benchmark_impact_initial_build
+);
 criterion_main!(benches);
--- a/backend/src/map_model.rs
+++ b/backend/src/map_model.rs
@@ -988,6 +988,7 @@ impl MapModel {
 
     // Lazily builds the router if needed.
     pub fn rebuild_router(&mut self, main_road_penalty: f64) {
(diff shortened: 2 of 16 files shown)


backend/src/map_model.rs
```diff
@@ -994,8 +994,11 @@ impl MapModel {
             .map(|r| r.main_road_penalty != main_road_penalty)
             .unwrap_or(true)
         {
-            let router_before_with_penalty =
-                Router::new(&self.router_input_before(), main_road_penalty);
+            let router_before_with_penalty = if self.router_before.main_road_penalty == main_road_penalty {
+                self.router_before.clone()
```
michaelkirk, line 998: This shaves off ~750ms from the 3.000 s initial load time if you haven't already loaded the router.

e.g. refresh the page, click "predict impact".

backend/src/od.rs
```diff
@@ -13,25 +12,36 @@ pub struct ZoneID(pub usize);
...
-            zone.x2 = (bbox.max().x * 100.0) as i64;
-            zone.y2 = (bbox.max().y * 100.0) as i64;
+            map.mercator.to_mercator_in_place(&mut zone.geometry);
+            prepared_zones.push(PreparedGeometry::from(&zone.geometry));
+        }
+
+        for road in &map.roads {
+            for (zone_idx, zone) in prepared_zones.iter().enumerate() {
+                if zone.relate(&road.linestring).is_intersects() {
+                    self.cached_zone_roads[zone_idx].push(road.id);
```
michaelkirk, line 38: Previously, to generate requests:

1. generate a random point within the zone.geometry bbox
2. do a `contains` check to make sure it was a valid point (else GOTO step 1)
3. Do a nearest neighbor query to get the road nearest that point

step 2 and 3 are both expensive. Now, we do this computation up front to see which roads are in each zone, and trip requests are made by grabbing random roads. It's not perfect but I think it's defensible, and it's faster, especially for the unsampled (more accurate) mode.

fast_sample: 713ms -> 666ms
all samples: 4459ms -> 2989ms

backend/benches/impact.rs
```diff
@@ -1,7 +1,25 @@
...
+fn benchmark_impact_initial_build(c: &mut Criterion) {
+    for fixture in [NeighbourhoodFixture::DUNDEE] {
+        let mut map = fixture.map_model().unwrap();
+        let fast_sample = false;
+
+        c.benchmark_group(fixture.savefile_name)
+            .sample_size(10)
+            .bench_function("predict impact - initial build", |b| {
+                b.iter(|| {
+                    map.rebuild_router(1.0);
```
michaelkirk, line 13: I'm not actually sure if this new bench is useful. TBH because there is some memoized state involved, it's a little tricky. Most of my measurements were done in the browser using the new console.log in Loading.svelte

web/src/common/Loading.svelte
```diff
@@ -0,0 +1,50 @@
...
+
+  let startTime: number | undefined = undefined;
+  let taskName: string | undefined = undefined;
+  $: {
+    if (loading && !startTime) {
+      // Start timing when loading begins
+      startTime = Date.now();
+      taskName = loading;
+    } else if (!loading && startTime) {
+      // Calculate and log duration when loading ends
```
michaelkirk, line 13: I copied this component from utils. The main change I made is that we now log the loading time whenever the loading banner is shown. It's not super precise, but since we're only ever showing the loading banner for relatively slow things, it's accurate enough to be helpful.

backend/src/od.rs
```diff
@@ -41,25 +51,42 @@ impl DemandModel {
...
+                return None;
+            }
+            let idx = rng.generate_range(0..slice.len());
+            Some(slice[idx])
+        }
+
+        // One sample will be made for this many trips when `fast_sample` is enabled.
+        // Increasing this will speed up fast_sample mode, but give less accurate results.
+        // Some OD pairs with less than this many trips might be skipped completely.
+        let trips_per_sampled_request = 10;
```
dabreegster, line 65: This could be much slower in areas with loads of trips, right, like maybe Edinburgh? Before with `fast_sample`, 5000 trips from zone1 to zone2 would just turn into one route. Now it'll do 500. I think you're right that this is the better approach -- the results from just sampling 1 were really biased and odd anyway.

web/src/common/Loading.svelte
dabreegster: Nice!

Eventually, I want to figure out how to plumb progress bar messages back from the WASM backend and really beef up this component. That's a long ways off though. This is a great step forward in the meantime.

dabreegster (approved):

backend/src/od.rs
dabreegster: I just read the commit message and comment more closely -- this new way is indeed more fair

michaelkirk: We can definitely crank this sample factor up if we want to speed things up further. In my own measurements, it's to a point where this part of the code is ~50% of the runtime, so you'll eventually see diminishing returns as the time spent building a router takes over.

Issue #300, opened 2025-04-03
diagonal filter rendered at incorrect rotation for some street angles


michaelkirk:
e.g. in this case it should be rotated counter clockwise another 20deg
<img width="111" alt="Image" src="https://github.com/user-attachments/assets/90c1be00-0f3c-4efd-953b-5328569b12cd" />

This is only a rendering issue - not a topological / routing one. 

IIRC the problem is that the rotational angle is calculated with reference to the angle between the two connected roads. When all the road angles are near 90deg, that works fine. But when you have an acute/obtuse angle, it gets wonky.

Instead, I think the reference angle should probably be the angle between the disconnected roads. Which is kind of obvious now that I type it out - since the purpose of the diagonal filter is to disconnect roads.

michaelkirk:
> Instead, I think the reference angle should probably be the angle between the disconnected roads. Which is kind of obvious now that I type it out - since the purpose of the diagonal filter is to disconnect roads.

After thinking about it another minute, it might be more complicated than that... we can see the diagonal does split the obtuse angle pretty well, just not the acute. Since the acute angle is going to be the hardest one to hit, maybe that's the most robust way:

1. Find the acutest angle in the intersection
2. Use that as your basis, the diagonal is either in the middle of the acute angle or 90deg from it, depending on the topological  orientation of the filter.

PR #325, opened 2025-04-10
FIX: Diagonal filter rotation for acute intersections

michaelkirk:
Previously it only looked good for intersections with ~90deg corners.

e.g. Adelaide Place and Law Street in Dundee

**before, followed by after, in the same video:**

https://github.com/user-attachments/assets/bb8771c9-d3ad-4aa8-a9e2-819976630820

### DIFF
--- a/backend/src/map_model.rs
+++ b/backend/src/map_model.rs
@@ -1,7 +1,7 @@
 use crate::boundary_stats::{ContextData, PreparedContextData};
 use crate::geo_helpers::{
-    aabb, angle_of_pt_on_line, bearing_from_endpoint, buffer_aabb, diagonal_bearing,
-    invert_multi_polygon, limit_angle, linestring_intersection,
+    aabb, angle_between_bearings, angle_of_pt_on_line, bearing_from_endpoint, buffer_aabb,
+    invert_multi_polygon, limit_angle, linestring_intersection, split_bearing,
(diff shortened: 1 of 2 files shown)


backend/src/map_model.rs
```diff
@@ -1188,12 +1188,31 @@ impl DiagonalFilter {
...
+            let bearing_b0 = bearing_from_endpoint(intersection.point, &road_b0.linestring);
+            let bearing_b1 = bearing_from_endpoint(intersection.point, &road_b1.linestring);
+
+            let angle_b0 = angle_between_bearings(bearing_a0, bearing_b0);
+            let angle_b1 = angle_between_bearings(bearing_a0, bearing_b1);
 
-        let bearing_1 = bearing_from_endpoint(intersection.point, &road_1.linestring);
-        let bearing_2 = bearing_from_endpoint(intersection.point, &road_2.linestring);
-        let angle = diagonal_bearing(bearing_1, bearing_2) as f32;
+            // Split the acute angle, since the obtuse angle has more tolerance
```
michaelkirk, line 1207: This is the key - rather than splitting between arbitrary roads, we make sure our split is relative to the most acute angle, which has the tightest tolerance.

dabreegster (approved):
FIXES #300

PR #326, opened 2025-04-11
Fix: tooltips off screen and introduce undesirable scrolling

michaelkirk:
Instead employ tippy.js

picocss tooltips were nice because they were completely declarative. However plain css wasn't powerful enough - we need more information about the current layout to position the tooltip visibly in some cases, which requires js.

Specific examples:

1. The "Add modal filter" button, had its tooltip mostly off the left edge of the screen.
2. Symmetrically, the "Redo" button, had its tooltip cropped off by the right edge of the left panel.
3. The "Redo" button and the "footnote" tooltips caused the left panel to be wider than necessary, even when they weren't being displayed, which introduced an undesirable (and very hard to track down!!!) horizontal scroll in the left panel.

The downside to this new non-declarative approach is that every component needs to initialize their tooltips by calling `initTooltips`. If a new component is rendered afterwards, its tooltips will not be initialized.

**before: **

Screenshot 2025-04-10 at 16 57 48

**after: **

Screenshot 2025-04-10 at 16 57 38

### DIFF
--- a/web/package-lock.json
+++ b/web/package-lock.json
@@ -15,7 +15,8 @@
         "route-snapper": "^0.4.9",
         "route-snapper-ts": "^0.0.8",
         "svelte-maplibre": "^0.9.14",
-        "svelte-utils": "github:a-b-street/svelte-utils"
+        "svelte-utils": "github:a-b-street/svelte-utils",
+        "tippy.js": "^6.3.7"
       },
       "devDependencies": {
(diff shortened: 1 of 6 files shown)


dabreegster (approved):
Tippy looks _much_ better. Having to be careful about initialization is not too bad in practice. Awesome!

Issue #324, opened 2025-04-10
Impossible one-way networks aren't detected


TFCx:
LTNv2 doesn't alert on impossible one-way configuration. A friend of mine is currently using the tool to simulate new circulation plans and that blocks him to rapidly evaluate if what he's doing contains a routing bug.

![Image](https://github.com/user-attachments/assets/f218a0c4-a8af-4302-8d1d-b020d5c89f32)

I searched for a similar issue but didn't find any. Hope i'm not mistaken.🤞

dabreegster:
There's not another issue for this yet, so thanks for filing. It might be a few weeks before I have time to figure out how to do this, but definitely is a good thing to add.

TFCx:
Just to confirm my friend's remark: I'm encountering this problem continuously right now ;)

mutricyl:
Using the "impact on one destination form" you can see problematic zones as ways are not drawn in wight anymore

<img width="810" height="524" alt="Image" src="https://github.com/user-attachments/assets/d30c3b5a-2d80-42d9-ba84-8dc285d12606" />

Likewise in the edit mode with "draw roads= worst shortcut" you are likely to have an issue when all shortcuts disappear when adding a one way.

PR #329, opened 2025-04-12
Take diagonal filters in account for cells

### LINKED ISSUE #322: Diagonal modal filter on intersection and background color
Dear LTN team 

I'm trying to add diagonal modal filter to isolate a small portion of a LTN. 
Starting from a small neighbourhood : 

Image

Then adding 2 x diagonal modal filter 

Image

It should create a specific area with a background color. But it's not calculating sub area as I expected. 

I can force this background color with modal filter applied on a road like this (but it's not the result I want) 

Image
 
Thanks for your help

&

=>  Congratulations on the new tool to reclassify main roads. It works well.   <=

dabreegster:
FIXES #322.

Here's a quick test case:
[diag cells-2025-04-12.geojson.txt](https://github.com/user-attachments/files/19721695/diag.cells-2025-04-12.geojson.txt)

Before, there's one big cell.
image

After, the cells reflect the possible movements.
image

### DIFF
--- a/backend/src/cells.rs
+++ b/backend/src/cells.rs
@@ -126,11 +126,11 @@ fn floodfill(map: &MapModel, start: RoadID, neighbourhood: &Neighbourhood) -> Ce
 
             for next in &map.get_i(i).roads {
                 let next_road = map.get_r(*next);
-                /*if let Some(ref filter) = map.get_i(i).modal_filter {
-                    if !filter.allows_turn(current.id, *next) {
+                if let Some(ref filter) = map.diagonal_filters.get(&i) {
+                    if !filter.allows_movement(&(current.id, *next)) {
                         continue;
                     }
-                }*/
+                }
                 if let Some(ref filter) = map.modal_filters.get(next) {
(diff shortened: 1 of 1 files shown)


backend/src/cells.rs
```diff
@@ -126,11 +126,11 @@ fn floodfill(map: &MapModel, start: RoadID, neighbourhood: &Neighbourhood) -> Ce
 
             for next in &map.get_i(i).roads {
                 let next_road = map.get_r(*next);
-                /*if let Some(ref filter) = map.get_i(i).modal_filter {
```
dabreegster, line 129: The code to do this was already here, probably backported from v1. I thought about using `allowed_movements_from` here as a larger cleanup, but it's a bigger change to do the floodfill by intersection, not by road. Maybe when we try #324, this could be useful

dabreegster:
Merging because it seems non-controversial, async feedback welcome

michaelkirk (commented):
LGTM - kind of fun to see the convergent evolution of the various versions.

Issue #191, opened 2025-03-06
Visualize "overall" metrics for merged boundary


michaelkirk:
Followup from https://github.com/a-b-street/ltn/pull/187#issue-2898877022

We currently show the aggregated metrics for your merged boundary:
<img width="298" alt="Image" src="https://github.com/user-attachments/assets/f8f4bdda-f49c-4661-a446-e133ce147e22" />

But it doesn't do a great job of comparison to the other potential boundaries. In effect, I want to communicate whether each metric's current value indicates if the boundary is a "good candidate" for an LTN

One specific proposal for how to surface this would be to annotate each metric's current value with our existing color schemes:

![Image](https://github.com/user-attachments/assets/3773b44f-68fa-4b60-979d-5a5909a911e3)

We currently have a uniform color schemes, but we might change that. In particular, today we were talking about leveraging a pre-existing color scheme for SIMD.

Another approach would be to have some kind of color neutral "progress" bar.

PR #330, opened 2025-04-19
Re-scale density metrics and add a combined metric

### LINKED ISSUE #303: POI and collision density scales unclear
https://en.wikipedia.org/wiki/Jenks_natural_breaks_optimization was one idea suggested to define the 5 different thresholds, per study area

dabreegster:
FIXES #303.

## Part 1

First up, this changes the limits for population, POI, and collision density to differ per study area. This is using https://docs.rs/ckmeans/latest/ckmeans/ to calculate 5 buckets. Cici originally recommended https://en.wikipedia.org/wiki/Jenks_natural_breaks_optimization, but in searching for crates to do this, I found `ckmeans`, which is much more polished and fast than the others I tried. Now things like population density can be seen for any area, ignoring any rural / urban differences. EDB population before is rather meaninglessly dense everywhere in the center:
image
Now the densest area is clear:
image

It's worth asking how we calculate these density buckets. We use the original census zone boundaries, **not** the severance boundaries we generate and initially display. The highest bucket threshold is not absolute; a user could draw a very small custom area around a bunch of collisions and wind up with a higher collision density than covered by the 5 buckets. The logic for later combining scores handles this. One practical implication is in Inverclyde. We don't see any severance boundary in the highest population density bucket:
image
But the layer on the right reveals the smaller zone achieving that density:
image

## Part 2

And by request from the last meeting, I've added a combined metric that turns each of the five metrics into a 1-5 score, then simply averages that. SIMD and car ownership are slightly special in that they are "reversed" relative to how they're shown currently. I'll revisit the ordering shown as part of #191, most likely.

image

I expect we may slightly tune the averaging or adjust wording after getting some feedback.


backend/src/boundary_stats.rs
```diff
@@ -203,6 +222,73 @@ pub struct PreparedPopulationZone {
     pub prepared_geometry: PreparedGeometry<'static, MultiPolygon>,
 }
 
+#[derive(Clone, Debug, Deserialize, Serialize)]
```
dabreegster, line 225: I moved the test section to the bottom, causing this diff

backend/src/boundary_stats.rs
```diff
@@ -203,6 +222,73 @@ pub struct PreparedPopulationZone {
...
+        (stats.number_stats19_collisions as f64) / stats.area_km2,
+        &context.metric_buckets.collision_density,
+    );
+    let pois = find_bucket(
+        (stats.number_pois as f64) / stats.area_km2,
+        &context.metric_buckets.poi_density,
+    );
+
+    // SIMD is [1, 100], with the [1, 20] group being the most important
+    let simd = find_bucket(100.0 - stats.simd, &[0, 20, 40, 60, 80, 100]);
```
dabreegster, line 258: Subtlety with these two

dabreegster:
I'll reupload the prioritization bin files before merging this

dabreegster:
I think there are some issues with the color scale for the combined 5, but I'm going to merge this in for feedback and continue working in https://github.com/a-b-street/ltn/tree/summarize_scores

michaelkirk (commented):
This LGTM.

I've proposed some simplifications here: https://github.com/a-b-street/ltn/pull/333

PR #331, opened 2025-04-19
When importing a new custom area, don't download and import twice

dabreegster:
I noticed something pretty awkward. Open up the dev console and create a new project from Overpass. Previously, we were downloading + importing, then immediately doing the same thing again due to calling `loadProject`. I've fixed this, verifying the console logs and making sure the new project created behaves correctly (saving changes, refreshing and loading it, etc)

### DIFF
--- a/web/src/title/NewProjectMode.svelte
+++ b/web/src/title/NewProjectMode.svelte
@@ -9,12 +9,13 @@
     appFocus,
     assetUrl,
     backend,
+    currentProjectID,
     map,
     mode,
     projectStorage,
   } from "../stores";
   import { Backend } from "../wasm";
-  import { loadProject } from "./loader";
+  import { afterProjectLoaded, loadProject } from "./loader";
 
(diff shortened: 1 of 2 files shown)


web/src/title/NewProjectMode.svelte
```diff
@@ -44,7 +45,11 @@
       );
 
       const projectID = $projectStorage.createProject($backend.toSavefile());
-      await loadProject(projectID);
+      // Calling loadProject would immediately re-fetch OSM data from Overpass
```
dabreegster, line 48: The first thing I tried was making `OverpassSelector` just hand us the boundary GJ and not actually load anything from Overpass itself. Then we could create an empty project and immediately load it, instead of having this second slightly weird code path. That approach would be cleaner, but I'm too lazy currently to go fix OverpassSelector and figure out if that kind of change also makes sense in other projects using it :)

Issue #255, opened 2025-03-20
CNT production deploy


michaelkirk:
Just marking that this needs to be done.

Some loose notes from today's discussion:

- [ ] procure tld for CNT (let's pretend it's cnt-example.org)
- [ ] CNT will be accessible from root at `https://cnt-example.org/` (not https://cnt-example.org/ltn/cnt)
  - [ ] Just to be clear, "global" will not be accessible from the production CNT deployment
- [ ] CDN
- [ ] certs?
- [ ] gzip assets

For the development site, it will be fine to have /ltn and /ltn/cnt urls at the current github domains to distinguish between the global|cnt flavor of the app.

dabreegster:
I've used Cloudflare R2 for storage and static file serving for other projects with some great success (minus trouble getting the gzip headers to work). I'll likely go with that too, unless other constraints come up or you have other recommendations.

dabreegster:
Literally just R2 with the default CDN in front of it. I think I've only used for static binary assets though, not HTML/CSS/JS, but I don't anticipate that's different

dabreegster:
I got something started with Cloudflare Pages. It builds automatically for the `prod_cnt` branch, which I plan to manually manage -- we may not to push changes from `main` there immediately. It's deployed to https://ltn-ynn.pages.dev/ successfully right now. Next I will:

- Register a domain
- Copy static files there (and get gzipping to work)

dabreegster:
I've registered cnt.scot through a third-party registrar, because Cloudflare doesn't support the .scot TLD. For my own future reference, I then:

1) Went through the "add custom domain" flow on Cloudflare
2) Changed the nameservers to Cloudflare through the registrar
3) DNS setup is "full" on cloudflare
4) Once cloudflare is managing the DNS, re-did the custom domain for Cloudflare Pages

The changes are now propagating

PR #332, opened 2025-04-20
Get cloudflare pages deployment working

dabreegster:
As part of #255, these're some changes I needed to make to get CNT deployed to cloudflare pages


.github/workflows/web.yml
```diff
@@ -31,7 +31,7 @@ jobs:
           cd web
           npm ci
           npm run wasm-release
-          npm run build --if-present
+          npm run build --if-present -- --base=/ltn/
```
dabreegster, line 34: On cloudflare, there's no leading directory, so only wire it up for GH pages

bin/cloud_build.sh
dabreegster, line 1: I originally pushed this file straight to main. The whole thing is new/unreviewed, but pretty straightforward

bin/cloud_build.sh
```diff
@@ -14,6 +15,6 @@ cd web
 npm ci
 npm run wasm-release
 npm run build --if-present
-# web/dist has the output
 
-# TODO Specialize for CNT
+# Overwrite the global version of the app with the CNT deployment
```
dabreegster, line 19: It really is that simple (after all your work to make AppFocus)

web/src/common/ModalFilterType.ts
dabreegster, line 1: I'm not happy how tedious and boilerplaety this is, but it appears necessary. I spent a few minutes unsuccessfully trying to change the vite env variable. But I don't want to hardcode a domain as the BASE_URL in there at all

PR #333, opened 2025-04-22
Clean up generate_prioritization

michaelkirk:
This isn't really a big deal, but a followup to #330 

By using the PopulationZoneInput:

1. We already have area, so we can skip building up and passing in a reduntant area_km2 vec.
2. We already have a prepared geometry.

And a minor pedantic improvement, by constructing the MetricBuckets before the ContextData, we can avoid initializing with a temporary bogus `Default` value.

### DIFF
--- a/data_prep/scotland/src/bin/generate_prioritization.rs
+++ b/data_prep/scotland/src/bin/generate_prioritization.rs
@@ -1,7 +1,7 @@
 use anyhow::{Context, Result};
 use backend::boundary_stats::{ContextData, MetricBuckets, POIKind, PopulationZone, POI};
 use data_prep::{PopulationZoneInput, StudyArea};
-use geo::{MultiPolygon, Point, PreparedGeometry, Relate};
+use geo::{MultiPolygon, Point, Relate};
 use serde::Deserialize;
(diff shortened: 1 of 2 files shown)


data_prep/scotland/src/bin/generate_prioritization.rs
```diff
@@ -235,61 +237,60 @@ impl SettlementsInput {
...
-    let population_zones = data
-        .population_zones
-        .iter()
-        .map(|zone| PreparedGeometry::from(&zone.geometry))
-        .collect::<Vec<_>>();
-
-    let mut collions_per_zone = Vec::new();
-    for zone in &population_zones {
+    let mut collisions_per_zone = Vec::new();
+    for zone in population_zone_inputs {
```
dabreegster, line 252: Just a note to self: before, the population zones passed in were pre-filtered to intersect the study area boundary. Now we pass all of them in. But the point-in-polygon test still behaves correctly, so works fine

dabreegster (approved):
Thank you, these cleanups are much appreciated! This is one consequence of using an editor without type completion; I lost track of what structs had which fields available

PR #334, opened 2025-04-23
Contextual layer tweaks

### LINKED ISSUE #147: context layers: population and density layers shouldn't be on the map at the same time
Just a small thing I noticed while using the app.

When both layers are shown, because they are translucent, you get color blending. Are we expecting the user to be able to interpolate across those two dimensions? It seems like it'd be pretty hard. I'd argue that enabling one should disable the other, like a radio toggle.

Some of the layers play nice together - e.g. POIs and density together are fine, but anything that's an "area" visualization might conflict with any other "area" visualization.

related #109 

https://github.com/user-attachments/assets/27e7c3d7-1f9b-41db-8e9c-b21af8eb2e08

dabreegster:
A few unrelated things:

1) Disable map rotation entirely. It's too easy to do accidentally and unclear how to undo, especially when you tilt. I don't think I've ever seen any reason to use it.

2) Rename a button per Cici's recommendation

3) Rename and group layers based on Cici's recommendation

4) Fix #147, keeping checkbox styling for consistency/simplicity

### DIFF
--- a/web/src/App.svelte
+++ b/web/src/App.svelte
@@ -98,6 +98,9 @@
 
   let map: Map | null = null;
   $: if (map) {
+    map.keyboard.disableRotation();
+    map.dragRotate.disable();
+    map.touchZoomRotate.disableRotation();
     mapStore.set(map);
   }
 
@@ -224,7 +227,7 @@
             },
           ]}
         >
-          <NavigationControl />
+          <NavigationControl showCompass={false} />
 
           {#if $backend}
             <Control position="top-left">
--- a/web/src/common/ModeLink.svelte
+++ b/web/src/common/ModeLink.svelte
@@ -5,14 +5,9 @@
   import { pageTitle } from "./navbar";
 
   export let mode: Mode;
-  export let afterLink: (() => void) | undefined = undefined;
 </script>
 
-<Link
-  on:click={() => {
-    $storedMode = mode;
-    afterLink && afterLink();
-  }}
+<Link on:click={() => ($storedMode = mode)}
   ><slot>
     <span class="page-title">{pageTitle(mode.mode)}</span>
   </slot>
--- a/web/src/context/CBD.svelte
+++ b/web/src/context/CBD.svelte
@@ -9,6 +9,7 @@
     SequentialLegend,
   } from "../common";
   import { assetUrl } from "../stores";
+  import RouteNetwork from "./RouteNetwork.svelte";
 
   // The NPT project bundles together a few layers into one pmtiles file, all
   // related to the Cycling By Design guidance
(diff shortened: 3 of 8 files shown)


web/src/common/ModeLink.svelte
```diff
@@ -5,14 +5,9 @@
   import { pageTitle } from "./navbar";
 
   export let mode: Mode;
-  export let afterLink: (() => void) | undefined = undefined;
```
dabreegster, line 8: Noticed this was unused

web/src/context/CBD.svelte
```diff
@@ -76,9 +77,16 @@
   </p>
 </ContextLayerButton>
 
-<ContextLayerButton bind:show={showExistingInfra} label="Cycle infrastructure">
+<RouteNetwork />
+
+<div class="layer-group">Other</div>
```
dabreegster, line 82: Slightly awkwardly, nest one of the layers in here, to achieve the desired grouping/ordering. One svelte component for both the map and controls remains useful, but means this kind of thing has to happen

web/src/context/Population.svelte
```diff
@@ -27,7 +27,16 @@
   let showCarOwnership = false;
 </script>
 
-<ContextLayerButton label="SIMD" bind:show={showSIMD}>
+<ContextLayerButton
+  label="SIMD"
+  bind:show={showSIMD}
+  onChange={() => {
```
dabreegster, line 33: One slightly disorienting thing is the position of the clicked button changing when the one above gets turned off. Slightly hard to see without the mouse showing up (ubuntu is still broken there)

[Screencast from 23-04-25 10:06:12.webm](https://github.com/user-attachments/assets/61ffad63-bb23-472f-b106-6424c1634540)

michaelkirk: Yeah, I see what you mean. It is a little disorienting, but I think the upside outweighs the downside.

web/src/App.svelte
```diff
@@ -98,6 +98,9 @@
 
   let map: Map | null = null;
   $: if (map) {
+    map.keyboard.disableRotation();
+    map.dragRotate.disable();
+    map.touchZoomRotate.disableRotation();
```
michaelkirk, line 103: Yes, thank you! I've only ever accidentally done this.

web/src/common/ModeLink.svelte
michaelkirk: Whoops! Some vestige of a bad idea of mine, no doubt.

web/src/context/CBD.svelte
michaelkirk: Would this awkwardness go away if we pulled the "traffic" layer and context button into a separate component? Then we could put the heading in ContextualLayers.svelte and put this new component after it.

Maybe there's a reason that's non-trivial. In any case, leaving it as is seems like no big deal.

michaelkirk (commented):
LGTM!

web/src/App.svelte
michaelkirk: For the record, I'll say that the map rotation is less of a problem for me. For one, I don't recall ever triggering it, and secondly, if I did, there is an obvious way to recover with the visible compass control. 

It's the "skew/perspective" control that always mystifies me, which has no obvious **visible** way to recover.

That said, I don't think I've ever really wanted to rotate the map either, so happy to see them all go away.

PR #336, opened 2025-04-25
Show simple progress bars for the neighbourhood stats summary

### LINKED ISSUE #191: Visualize "overall" metrics for merged boundary

dabreegster:
FIXES #191.

[Screencast from 25-04-25 17:57:01.webm](https://github.com/user-attachments/assets/e9130ba5-9f62-46b3-a2c0-7af8c659a0b3)

I've gone with a "progress bar" that shows the bucket colors. Cici and Angus liked this enough.

Note car ownership and SIMD are still reversed relative to the other metrics; I'll fix that separately. Showing the colors makes things explicit, in the meantime.

### DIFF
--- a/web/src/common/MetricProgress.svelte
+++ b/web/src/common/MetricProgress.svelte
@@ -1,0 +1,43 @@
+<script lang="ts">
+  export let colorScale;
+  export let limits: number[];
+  export let value: number;
+
+  $: bucketIdx = calculateBucket(limits, value);
+
+  function calculateBucket(limits: number[], value: number): number {
+    // Note the value might exceed the highest limit; it winds up in the last bucket if so
+    let idx = 0;
+    for (let limit of limits.slice(1, -1)) {
+      if (value < limit) {
+        break;
+      }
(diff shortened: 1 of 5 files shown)


michaelkirk:
Oh this is really nice! 

I've made two proposed amendments here: https://github.com/a-b-street/ltn/pull/337

Feel free to incorporate either, all, or neither.

michaelkirk (approved):

PR #337, opened 2025-04-25
summarize scores design tweaks

michaelkirk:
Some suggested addendums to #336 

## before

Screenshot 2025-04-25 at 10 40 29

## after

1. Show composite score legend as "buckets" rather than sequential.

Screenshot 2025-04-25 at 11 03 22

---

2. Replaced red border with color-blind friendly indicator. I actually thought it was a black border at first, and I was losing it in the darker blue colors. There was also a minor jitter when moving the border from cell to cell as you add/remove areas. It wasn't noticable if the border moved to an adjacent category, but if you "leapfrogged" over a category (e.g. simd 1 to simd 3) there was a little unfortunate width wiggling.

Screenshot 2025-04-25 at 11 03 13

### DIFF
--- a/web/src/common/MetricProgress.svelte
+++ b/web/src/common/MetricProgress.svelte
@@ -1,4 +1,6 @@
 <script lang="ts">
+  import { ArrowUp } from "lucide-svelte";
+
   export let colorScale;
   export let limits: number[];
   export let value: number;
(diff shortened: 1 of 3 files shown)


dabreegster:
Erm, I thought this was a PR against main, let me fix this

dabreegster:
Thank you for both improvements, they're great! I've pushed both commits to main

Issue #192, opened 2025-03-06
Use conventional coloring for SIMD


michaelkirk:
Currently, we have a common color ramp we use for all our metrics.

The SIMD metric has a [well known tool](https://simd.scot/#/simd2020/BTTTFTT/9/-4.0000/55.9000/) which has its own color ramp. 

<img width="538" alt="Image" src="https://github.com/user-attachments/assets/7d456f07-0e87-498d-b3bc-d16884068c87" />

If people have internalized that color ramp, we might potentially confuse them by showing different colors. In particular our blue means the opposite of the blue at  https://simd.scot/#/simd2020/BTTTFTT/9/-4.0000/55.9000/

On the other hand, our existing color ramp is known to be designed as color-blind friendly.

This choice might also affect our thinking on #191

michaelkirk:
A couple complications here:

This conventional coloring is a 10 point scale, not a 5 point scale like we use for all our other metrics.
- We could combine their buckets so that there are only 5 (though the middle one will be tough)
- Alternatively, we could introduce additional buckets to our UI to match the categories in this given scale. I think that should be relatively straight forward. Though if the legend is too crowded for 10 categories, we might have two break it into two legends like they've done: the first for 0-50%, and the second for 50-100%

Also worth noting that their "least deprived" color is unfortunately close to the color we use in our other metrics to indicate "likely a better candidate for LTN". So it's sort of opposite of how we use that color for all our other metrics.

I can do this work, but wanted to make sure the cost/benefit makes sense to prioritize this.

michaelkirk:
Hi @congying-hu! 

Let me try to restate my previous question: The example colors we were asked to adopt are on a color scale with 10 colors. 

I'm fine to continue using 5 categories (quintiles) as we are now, but to color 5 categories, I need 5 colors, not 10. What should those 5 colors be?

michaelkirk:
> I think we can keep SIMD color scheme and reasonably choose 5 colours.

@congying-hu - please send me the hex codes for the 5 colors you'd like to use for SIMD and I'll plug them in.

> so that deepest red always indicate worse situation consistently:

Like this?

https://github.com/user-attachments/assets/e95add41-f4ff-4fb5-91f8-1d1ead19d5a7

PR #338, opened 2025-04-26
Make metric color scales consistent

dabreegster:
SIMD and car ownership were previously flipped, showing the most important category on the left. This PR makes everything consistent:

image

Also, I've been having trouble seeing the dashed outline of the current boundary selection against the new colors. I've tried red/black there instead of white/black -- what do you think? But I've left pick neighbourhood alone.

The color scales might change more with #192, but this at least gets things in the same order.


web/src/common/colors.ts
```diff
@@ -32,6 +34,8 @@ export let stats19ColorScale = commonQuintileColorScale.toReversed();
 
 export let poiColorScale = commonQuintileColorScale.toReversed();
 
+// Notice parts of the UI reverse these, to treat lowest ownership as the most
```
dabreegster, line 37: I wasn't sure if it'd be cleaner to reverse here for most color scales / legends, and then un-reverse when we do something like `makeRamp` or `MetricBucket` that expects ascending limits.

web/src/common/colors.ts
```diff
@@ -17,8 +17,10 @@ let commonQuintileColorScale = [
   "#CDE594",
 ];
 
+// The percentiles are [1, 100]. The 5 colors cover 4 each. Notice parts of the
```
michaelkirk, line 20: > The 5 colors cover 4 each.

I know this comment was pre-existing, but I don't understand what it means.

web/src/common/colors.ts
michaelkirk: It's a bit complicated, but I think it's fine.

michaelkirk (commented):
Personally, I find it harder to see the new border style. It mostly looks like black on black to me.

michaelkirk (approved):
Whoops! I meant to approve this.

web/src/AddNeighbourhoodMode.svelte
```diff
@@ -401,15 +401,15 @@
...
             "line-dasharray": [2, 2],
           }}
         />
         <LineLayer
           {...layerId("neighbourhood-boundaries-selected-outline-base", false)}
           paint={{
-            "line-color": "white",
+            "line-color": "black",
             "line-width": 8,
             "line-opacity": 0.7,
```
michaelkirk, line 414: If you want to spend more time on this... I think a more color blind friendly approach would be to revert to the black dashes on the white line, but increase the line-opacity of the white.

dabreegster: Hmm, 0.3 opacity on the white looks like this:
image
Because we use black for main roads, at low zooms, it all just blurs together a bit. I'll revert this last commit and keep trying some more ideas later. Thanks for the feedback!

web/src/common/colors.ts
dabreegster: I deleted this sentence entirely. I also don't know what I meant anymore; I think I was trying to talk myself through the bucketed legend or something.

web/src/AddNeighbourhoodMode.svelte
michaelkirk: > Because we use black for main roads, at low zooms, it all just blurs together a bit.

Yes, exactly! That's why I said *increase* the opacity, not decrease.

dabreegster: Enthusiasm: 9/10
Reading comprehension: 2/10

image
image
This looks great, I'm going to commit it, and happy to adjust later

PR #339, opened 2025-04-26
Heatmap for collisions

dabreegster:
The sea of stats19 dots (with popups!) at low zoom isn't helpful. Switch to a heatmap instead. I've mostly kept maplibre's default heatmap styling, but adjusted the clustering slightly to look decent in both sparse and dense areas. Happy to do more tuning though!

image
image

(Videos are hitting the 10MB limit too quickly, with so much on the screen)

### DIFF
--- a/web/src/common/ContextLayerButton.svelte
+++ b/web/src/common/ContextLayerButton.svelte
@@ -8,7 +8,6 @@
 
 <button
   class="context-control"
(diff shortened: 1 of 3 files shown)


web/src/common/ContextLayerButton.svelte
dabreegster, line 1: Unrelated, just lifting this out, in prep for trying to shrink / re-style the layer menu a bit

michaelkirk (approved):

PR #340, opened 2025-04-29
Cleanly handle one error case when drawing boundaries

### LINKED ISSUE #257: debug_assert triggered when drawing new neighborhood boundary
Followup from https://github.com/a-b-street/ltn/pull/246#pullrequestreview-2702087008

> I managed to trip the debug assert in make_polygon_valid about multipolygon results. I'll get a good repro in Dundee and file an issue. I think because we now call this method much more frequently while dragging around, it was easier to hit. I think in the short-term, changing this to return a Result instead of panic and handling it cleanly on the frontend should help.

I'd want to see what the repro is before disabling the assert. It's a debug assert, so it won't be triggered in the release build, and it might indicate a bug that we can better address.

dabreegster:
FIXES #257 
[Screencast from 29-04-25 09:58:06.webm](https://github.com/user-attachments/assets/bb0f61b3-0ccd-4a8a-b4ee-f3122e71f23f)

The behavior in debug mode feels correct to me here. The 3 points don't form a valid area for this odd case, because they just double back to try and connect. It feels intuitive to me to have the left sidebar reflect that I haven't made a valid area yet.

### DIFF
--- a/backend/src/geo_helpers/mod.rs
+++ b/backend/src/geo_helpers/mod.rs
@@ -4,6 +4,7 @@ pub use slice_nearest_boundary::SliceNearestFrechetBoundary;
 mod roads_along_line;
 pub use roads_along_line::roads_along_line;
 
+use anyhow::Result;
 use geo::line_measures::InterpolatableLine;
 use geo::{
     BooleanOps, BoundingRect, Buffer, Contains, Coord, Distance, Euclidean, Intersects, Length,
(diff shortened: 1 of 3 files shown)


backend/src/geo_helpers/mod.rs
```diff
@@ -208,23 +209,22 @@ fn euclidean_destination_coord(pt: Coord, angle_degs: f64, dist_away_m: f64) ->
...
-    debug_assert!(
-        valid_multipolygon.0.len() == 1,
-        "MultiPolygon not handle yet"
-    );
+    if valid_multipolygon.0.len() != 1 {
+        bail!("MultiPolygon not handled yet");
+    }
 
     let Some(valid_polygon) = valid_multipolygon.0.pop() else {
         debug_assert!(false, "empty valid polygon not handled yet");
```
dabreegster, line 222: I left in the other two debug asserts, and have never tripped them yet

backend/src/lib.rs
```diff
@@ -192,8 +192,7 @@ impl LTN {
             .map(|feature| {
                 let mut polygon = Polygon::try_from(feature).map_err(err_to_js)?;
                 self.map.mercator.to_mercator_in_place(&mut polygon);
-                make_polygon_valid(&mut polygon);
-                Ok(polygon)
+                make_polygon_valid(&polygon).map_err(err_to_js)
```
dabreegster, line 195: Big change here! Previously we were calling `make_polygon_valid` but throwing away the result; it doesn't modify anything in-place. The only effect was to possibly crash in debug mode. I haven't noticed any behavioral differences now that this is using the result. Was the prior behavior deliberate?

michaelkirk: Given the confusing method name and the mutable references, it's almost surely an error on my part. 

I think the intention was, in the case of a Multipolygon, since we don't handle MultiPolygons here, to continue on with one of its inner Polygons.

We'll no longer do that at the other `make_polygon_valid` callsite. Maybe that's fine and not something I should have been trying to avoid.

michaelkirk (approved):

backend/src/geo_helpers/mod.rs
```diff
@@ -208,23 +209,22 @@ fn euclidean_destination_coord(pt: Coord, angle_degs: f64, dist_away_m: f64) ->
 
 /// Attempts to make the input polygon valid by union-ing it with itself.
 ///
-/// It hasn't been thoroughly tested.
-pub fn make_polygon_valid(polygon: &Polygon) -> Polygon {
+/// It hasn't been thoroughly tested. Some edge cases will panic only when built in debug mode.
+pub fn make_polygon_valid(polygon: &Polygon) -> Result<Polygon> {
     let mut valid_multipolygon = polygon.union(polygon);
 
     // I don't think we should get more than one piece back for any sane input, but we'll see...
```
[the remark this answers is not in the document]
dabreegster: Done

PR #344, opened 2025-05-02
Add an area prioritization metric

### LINKED ISSUE #193: boundary merging ui: Help user "right size" their LTN
Given some guidelines on the appropriate size of an LTN, we could surface this to the user to help them grow/shrink their LTN.

Since we're showing the area of the assembled boundary:

Image

There are a couple ways we could indicate "too small" or "too big".

This might relate to  #191. Even though Area isn't a "prioritization" metric in the same way as the others — we can't filter by it — we might still be able to utilize similar design treatments for right sizing Area in the stats table.

Another thought is some kind of highlight in the "Add to area" section if the Area is too small, or update the copy to reflect "too big".

dabreegster:
FIXES #193, at least in spirit.

We decided to include an area metric after all, but not to include it in the combined score. We're using limits and colors from page 7 of https://content.tfl.gov.uk/lsp-app-six-b-strategic-neighbourhoods-analysis-v1.pdf, because there's no Scottish guidance.

image

### DIFF
--- a/web/src/common/colors.ts
+++ b/web/src/common/colors.ts
@@ -26,8 +26,15 @@ export let populationDensityColorScale = commonQuintileColorScale.toReversed();
 
 export let demandColorScale = commonQuintileColorScale.toReversed();
 
-export let areaLimits = [0.0, 0.3, 0.6, 1.0, 1.5, 2.0];
-export let areaColorScale = commonQuintileColorScale.toReversed();
+// From page 7 of https://content.tfl.gov.uk/lsp-app-six-b-strategic-neighbourhoods-analysis-v1.pdf, except removing the smallest bucket to make five
+export let areaLimits = [0.0, 0.25, 0.5, 1, 1.5, 2.0];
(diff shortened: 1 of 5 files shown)


web/src/common/colors.ts
```diff
@@ -26,8 +26,15 @@ export let populationDensityColorScale = commonQuintileColorScale.toReversed();
 
 export let demandColorScale = commonQuintileColorScale.toReversed();
 
-export let areaLimits = [0.0, 0.3, 0.6, 1.0, 1.5, 2.0];
```
dabreegster, line 29: We still had remnants of this leftover

dabreegster:
Feedback from Cici:

- the color scale maybe isn't clear, that the best values are in the middle
- maybe it shouldn't be in the table at the bottom for a custom area, since it's not part of the combined score. (Dustin: But I think it's useful for the "right-sizing" exercise)

If you have any opinions, might adjust, but will default to just continuing as-is

Issue #345, opened 2025-05-02
Performances became worse?


TFCx:
I'm working on a very large zone (23km²) and I think the performances became worse (in comparison to 2/3 months ago). Any modification needs a 1 minute update. It's a bit frustrating because I could imagine doing multiple modifications without the need for intermediate visualization. I tried turning off the computing of entrances/exits and switching to edits road coloring (which should not implicate any computation ?) but I doesn't speed up the process.

https://a-b-street.github.io/ltn/?project=b4bb49da-126e-477a-8f94-f976261388b8#13.12/43.60493/3.87743

![Image](https://github.com/user-attachments/assets/34aadefc-d95e-45a9-a910-0df6f8e6858a)

It's not blocking since I could focus on a pre-defined neighborhood but it complicates a bit the workflow for me.

dabreegster:
Thanks for reporting. Mind exporting the project file and attaching (maybe you have to rename to .txt to trick Github)? This is kind of a known issue; even 2-3 months ago, I was trying to use for a large English city centre and also felt how painfully slow things were. Now that the feature development has settled down, a pass at performance for these large cases would be great. It's still a few more weeks before I'll able to prioritize this, though.

dabreegster:
How are you finding the sea of icons at this scale and zoom by the way? It looks very noisy in your screenshot. You can disable existing filters in the layers on the right, but maybe we could also reconsider styling at mid/low zooms like this.

TFCx:
[Montpellier Plan 1-2025-05-02.geojson.txt](https://github.com/user-attachments/files/20013361/Montpellier.Plan.1-2025-05-02.geojson.txt) (the project only has like 2/3 user modifications).

TFCx:
Sea of icons ➡ Yes, It's noisy. I would try to scale the icons down when in an intermediate zooming (like on the screenshot) and use about the same diameter at the road width. Just to express that "there is some kind of flow modification here" even if it's a not completly readable. (You could also image that hovering over on of those icons would scale the icon temporarily to understand what's under your mouse). Of course, when zooming in, I would still use the bigger than road icons (= current situation).

dabreegster:
[...]

PR #353, opened 2025-05-05
Don't calculate shortcuts unless we're animating

dabreegster:
A simple start to #345 -- roughly halve the time after every edit, unless the user has opted into animating shortcuts. For huge/slow areas, I don't believe anybody reasonably would. So should be an immediately noticeable boost.


michaelkirk (commented):
LGTM

PR #354, opened 2025-05-07
Refine shortcut definition

### LINKED ISSUE #124: When counting shortcuts, shortcut must cross interior road
When perimeter roads are editable, think through if the shortcuts shown make sense

dabreegster:
FIXES #124. Addresses feedback from #352.

First, this reverts part of #352, to account for different directions of entrance. The argument about turn restrictions is convincing.

Second, this fixes the case of shortcuts incorrectly crossing right over a main road:
image

------

**Heads up** I'm not sure about merging this yet. There's a test diff in bristol_west that looks problematic to me. Before, this is a shortcut:
image
With this PR, there are no shortcuts on that westbound one-way road:
image

This looks incorrect to me; there should be a shortcut there. I think the difference is caused by the west end of the shortcut:
image
This is a weird case of a three-way intersection between one main road and two interior roads, at https://www.openstreetmap.org/way/3567399#map=19/51.454404/-2.568893. The main road is one-way, in an incompatible direction with the westbound one-way that's part of the shortcut. So actually yeah, the check for crossing through border intersections is maybe insufficient. In both cases (the shortcut that should not be counted, and this one that should), each interior road on either side of the border intersection is in a different cell. So I'm not sure how to distinguish these cases yet. This may be an edge case we're willing to sacrifice, for the more common case of crossing a main road.


dabreegster:
Just thinking aloud: the distinction between the two cases is whether the movement between interior roads crosses over a main road. In the first example
image
We really do cross the main road at that intersection.

In this second case, we do not. The two interior roads are adjacent in the clockwiseness. I'll refine this check and reopen.

dabreegster:
https://github.com/a-b-street/ltn/issues/345#issuecomment-2862358828
I'll keep this draft until I also try out this idea.

PR #355, opened 2025-05-08
Dramatically speed up how shortcuts are found, using cells to restrict

### LINKED ISSUE #124: When counting shortcuts, shortcut must cross interior road

dabreegster:
candidates

FIXES #124 and makes huge progress on #345.

Previously, we used every pair of border intersections in the neighbourhood as candidates for shortcuts. This was wasteful, because if we try to find a route between different cells, then by the definition of cells, the route has to cross or go along a main road. In one example larger neighbourhood, there were 145,000 wasted routes calculated, for only 1,000 valid ones. Calculating cells is cheap and we do it anyway, so we can simplify the shortcut calculation and only look for shortcuts within one cell at a time.

#354 proposed a check for border intersections at every intersection along a candidate route. We don't need this here, because we're never trying to cross over a main road now with the requests in the first place.

There are two test diffs. In Bristol, the shortcuts before this PR look like this:
image
With this PR, we lose the shortcuts on Victoria Avenue:
image
This is caused by a funny small cell on the western edge:
image
I think this is acceptable, because this is a case where the automatic main road detection is a little wrong. When a one-way main road points into a neighbourhood like this, it feels to me more like an interior road; the border should be a bit further back. If I manually reclassify and make it look correct:
image
Then the shortcut on Victoria Avenue works fine:
image

Strasbourg has a change only to the number of shortcuts along the eastern edge. Two shortcuts get perfectly repeated; they are totally identical. One is:
image
The other is:
image
I'm not entirely sure why, but the eastern edge where both of these duplicates start looks odd to me, with an interior road in between two main roads. I'm not sure best practice is for neighbourhood definitions to include both of the parallel dual carriageway pieces, so we could even consider editing this test definition. (Maybe the same for Bristol too?)
image

In terms of speed impact, the benchmark even for the pretty simple test areas gets much better, because still pruning by cells is useful, even if there's only "one" main road loop.
```
shortcuts in bristol_east
                        time:   [1.9851 ms 2.1903 ms 2.4327 ms]
                        change: [-28.750% -24.053% -19.337%] (p = 0.00 < 0.05)
                        Performance has improved.

shortcuts in bristol_west
                        time:   [6.0424 ms 6.2505 ms 6.5135 ms]
                        change: [-50.039% -47.191% -44.010%] (p = 0.00 < 0.05)
                        Performance has improved.

shortcuts in strasbourg time:   [2.5492 ms 2.5572 ms 2.5660 ms]
                        change: [-33.187% -31.986% -30.802%] (p = 0.00 < 0.05)
                        Performance has improved.
```

And for the huge city centre areas, I haven't measured in release mode yet, but just on dev mode, one of my examples goes from 3-4s down to 200ms! I'll try release mode on the #345 example, but I bet that 6.2s example is subsecond now.


backend/src/cells.rs
```diff
@@ -10,13 +10,13 @@ pub struct Cell {
     /// between two cells, and the PercentInterval indicates the split.
     pub roads: BTreeMap<RoadID, PercentInterval>,
     /// Intersections where this cell touches the boundary of the neighbourhood.
-    pub borders: BTreeSet<IntersectionID>,
+    pub border_intersections: BTreeSet<IntersectionID>,
```
dabreegster, line 13: Just renaming for consistency

backend/src/lib.rs
```diff
@@ -434,8 +434,16 @@ impl LTN {
 
     #[wasm_bindgen(js_name = getShortcutsCrossingRoad)]
     pub fn get_shortcuts_crossing_road(&self, road: usize) -> Result<String, JsValue> {
+        let derived = self
```
dabreegster, line 437: These two calls were wastefully recalculating shortcuts from scratch, even though we already calculate and store them. `derived` should never be `None` past the constructor, and every mutation calls `after_edit`, so this is safe

backend/src/shortcuts.rs
```diff
@@ -18,107 +18,113 @@ pub struct Path {
...
-            let start_intersection = map.get_i(*start_i);
-            for start_r in &start_intersection.roads {
-                // It's not a "shortcut" unless it starts outside the interior and cuts through the interior.
-                if !neighbourhood.main_roads.contains(start_r) {
-                    continue;
-                }
-                for end_i in &neighbourhood.border_intersections {
-                    if start_i == end_i {
+
+        // A shortcut is contained within a cell, because cells can't cross modal filters or main
```
dabreegster, line 29: Ignoring whitespace in the diff is useful. Main change is here.

backend/src/shortcuts.rs
```diff
@@ -18,107 +18,113 @@ pub struct Path {
...
-                        let direct_length = match map.router_before.route_from_roads(
-                            &router_input_before,
-                            *start_r,
-                            *end_r,
-                        ) {
-                            Some(route) => Euclidean.length(&route.to_linestring(map)),
-                            None => {
-                                warn!("Found a shortcut from {start_r} to {end_r}, but not a route using the whole map");
-                                shortcut_length
+                            // Only increase count_per_road after verifying above that the shortcut
```
dabreegster, line 101: Added a comment here to explain why we need two loops

dabreegster:
image
In this massive area in release mode, total update time is down to 251ms: Finding cells took 3ms, rendering cells took 166ms, finding shortcuts took 82ms

backend/src/shortcuts.rs
```diff
@@ -18,107 +18,113 @@ pub struct Path {
...
-                    let end_intersection = map.get_i(*end_i);
-                    'next_road: for end_r in &end_intersection.roads {
-                        // It's not a "shortcut" unless it ends outside the interior after cutting through.
-                        if !neighbourhood.main_roads.contains(end_r) {
-                            continue;
-                        }
-
-                        // It's possible to repeat the route request when the same road is
-                        // connected to multiple border intersections
-                        if checked_routes.contains(&(*start_r, *end_r)) {
```
michaelkirk, line 49: To make sure I'm understanding: This `checked_routes.contains` call was removed because it is redundant, now that we only considers trips within a single cell.

Is that right?

michaelkirk (approved):
Good work!

michaelkirk:
> With this PR, we lose the shortcuts on Victoria Avenue:
>
> This is caused by a funny small cell on the western edge:

I agree that it's a strange cell, and I agree that it's acceptable that the cell exists given the input data.

That said, I don't really understand why this change would result in a big swing in shortcuts for this case.

backend/src/shortcuts.rs
dabreegster: I removed it because of the reasoning in https://github.com/a-b-street/ltn/issues/124#issuecomment-2859379071 actually -- the same request could be formed from multiple directions, and to let turn restrictions at the border have an effect of shortcuts, we don't want to do this deduping.

Issue #311, opened 2025-04-04
Improve perf for initially loading


dabreegster:
- [x] Serve gzipped files, #239 
- [x] Consider adding a progress bar for loading the file(s)
- [ ] Measure how long it's taking to create a MapModel from osm.pbf, compared to pre-building and deserializing bincoded MapModel files

PR #356, opened 2025-05-08
Enhance the initial loading screen to show download progress and sepa…

dabreegster:
…rately indicate parsing/setup

Part of #311. Give the user more info about what's happening during initial setup, for all cases of CNT/global mode, downloading from overpass vs a pre-clipped file, downloading vs CPU-bound setup.

It's a nicer view on slow internet, but here's one example:

[Screencast from 08-05-25 14:18:03.webm](https://github.com/user-attachments/assets/5d92255c-f15a-4a91-871e-f497ef0db5b9)

### DIFF
--- a/web/src/AddNeighbourhoodMode.svelte
+++ b/web/src/AddNeighbourhoodMode.svelte
@@ -20,6 +20,7 @@
     ModeLink,
     pageTitle,
     prettyPrintPercent,
+    refreshLoadingScreen,
   } from "./common";
   import AreaControls from "./common/draw_area/AreaControls.svelte";
   import { type Waypoint } from "./common/draw_area/stores";
@@ -88,17 +89,7 @@
   }
 
   onMount(async () => {
-    // Make sure the loading indicator has rendered before we generate boundaries.
-    //
-    // NOTE: I thought this is what `svelte.tick` was for, but it doesn't suffice.
-    //     await tick();
-    await new Promise((resolve) => {
-      requestAnimationFrame(() => {
-        // Waiting for one frame also doesn't seem to be enough, so we wait for two. *shrug*
-        requestAnimationFrame(resolve);
-      });
-    });
-
+    await refreshLoadingScreen();
     generatedBoundaries = $backend!.generatedBoundaries();
     loading = "";
   });
@@ -237,6 +228,7 @@
 </script>
 
 <Loading {loading} />
+
 <SplitComponent>
   <div slot="top">
     <nav aria-label="breadcrumb">
--- a/web/src/CntChooseArea.svelte
+++ b/web/src/CntChooseArea.svelte
@@ -14,8 +14,6 @@
   import { mode, projectStorage } from "./stores";
   import { loadProject } from "./title/loader";
 
-  export let activityIndicatorText: string;
-
   let gj: FeatureCollection<
     Polygon | MultiPolygon,
     { kind: "LAD"; name: string }
@@ -65,7 +63,6 @@
         );
         continue;
       }
-      activityIndicatorText = `Loading pre-clipped OSM area ${prettyPrintStudyAreaName(studyAreaName)}`;
       try {
         let projectID = $projectStorage.createEmptyProject(
           projectName,
@@ -77,7 +74,6 @@
       } catch (e) {
         window.alert(e);
       }
-      activityIndicatorText = "";
     }
   }
 </script>
--- a/web/src/common/Loading.svelte
+++ b/web/src/common/Loading.svelte
@@ -1,6 +1,6 @@
 <script lang="ts">
-  export let loading;
-  export let progress = null;
+  export let loading: string;
+  export let progress: number | null = null;
 
   let startTime: number | undefined = undefined;
   let taskName: string | undefined = undefined;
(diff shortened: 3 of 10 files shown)


web/src/AddNeighbourhoodMode.svelte
```diff
@@ -88,17 +89,7 @@
...
-    // NOTE: I thought this is what `svelte.tick` was for, but it doesn't suffice.
-    //     await tick();
-    await new Promise((resolve) => {
-      requestAnimationFrame(() => {
-        // Waiting for one frame also doesn't seem to be enough, so we wait for two. *shrug*
-        requestAnimationFrame(resolve);
-      });
-    });
-
+    await refreshLoadingScreen();
```
dabreegster, line 92: Also refactored this pattern

web/src/CntChooseArea.svelte
```diff
@@ -14,8 +14,6 @@
   import { mode, projectStorage } from "./stores";
   import { loadProject } from "./title/loader";
 
-  export let activityIndicatorText: string;
```
dabreegster, line 17: Now `loadProject` handles it

web/src/common/Loading.svelte
```diff
@@ -20,13 +20,19 @@
...
+      {loading}
 
-    {#if progress != null}
-      <div>
-        <progress value={progress} style:width="100%" />
-      </div>
-    {/if}
+      {#if progress != null}
+        <div>
+          {#if progress == 100}
```
dabreegster, line 28: When progress is 100 (either because it's set explicitly) or it reaches it from a download, switch to the bounce-back-and-forth-constantly style that just means "something is happening"

web/src/title/NewProjectMode.svelte
```diff
@@ -70,13 +75,14 @@
...
       return;
     }
-    loading = `Loading pre-clipped OSM area ${example}`;
     await loadProject(projectID);
     $mode = { mode: "add-neighbourhood" };
-    loading = "";
   }
 </script>
 
+<Loading {loading} progress={100} />
```
dabreegster, line 83: Slightly weird, but two cases are possible from here -- directly start pulling from overpass and then create a project, or create a project file and load from it the normal way

web/src/title/loader.ts
```diff
@@ -19,8 +20,12 @@ import {
 } from "../stores";
 import { Backend } from "../wasm";
 
+export let loadingMessage = writable("");
+export let loadingProgress: Writable<number | null> = writable(null);
+
 /**
- * Loads an existing project by its ID
+ * Loads an existing project by its ID. Changes loadingMessage and
```
dabreegster, line 27: Updating these stores was the most sane way I could think of to plumb updates out

web/src/title/loader.ts
```diff
@@ -29,6 +34,11 @@ export async function loadProject(projectID: ProjectID) {
     let { osmBuffer, demandBuffer, contextDataBuffer, boundary } =
       await getInputFiles(project);
     console.timeEnd("get input files");
+    loadingMessage.set("Download finished, setting up project");
+    // TODO The animation won't work, because we block the UI thread below
```
dabreegster, line 38: When we do any CPU-bound step, it blocks the UI thread, so the browser's progress animation stops. Eventually we might want to make everything run async in web workers, but not important for now

michaelkirk (commented):
I'm not sure if this is a safari specific thing, but the text is clipped:

Screenshot 2025-05-08 at 12 22 35

Also, presumably because the download happens quickly, the first thing I see is "download finished" it's a little confusing since, from my perspective, the download never happened.

A checklist style UI might be more helpful:

---
- [x] Downloading map data
- [ ] Setting up project
---

Also, the "frozen" progress bar, for me anyway, is frozen pretty much most of the time. I think a generic activity indicator, animated by css, which doesn't communicate anything about percentage complete would likely be more helpful, and maybe could avoid the problem of freezing.

All that said, that could be follow work for a later day.

dabreegster:
I've filed #361 to follow up on these, thanks!

> I'm not sure if this is a safari specific thing, but the text is clipped:

Strange, that's where the bouncing progress should go. Maybe the blocking WASM call somehow interrupted the browser mid-rerender, though that's not how I thought animation frames worked.

> A checklist style UI might be more helpful

Ooh, I really like this idea! It'll take some more thought for a structure to represent this, and could maybe also be useful for plumbing progress info back from WASM (the old abstreet `Timer`), but I'll leave this for later.

>  I think a generic activity indicator, animated by css, which doesn't communicate anything about percentage complete would likely be more helpful

That's what should appear in your text clipping screenshot, just a `<progress />` without a value and max. On Firefox, it bounces back and forth like a generic spinner. But because we do a big blocking call right from the UI thread, it still freezes. Only way to avoid that is to use web workers, I think.

michaelkirk:
>> I think a generic activity indicator, animated by css, which doesn't communicate anything about percentage complete would likely be more helpful
> That's what should appear in your text clipping screenshot, just a <progress /> without a value and max. On Firefox, it bounces back and forth like a generic spinner. But because we do a big blocking call right from the UI thread, it still freezes. Only way to avoid that is to use web workers, I think.

Currently it looks like we set the % complete  as width in javascript (as part of rendering in svelte)..

```
<progress value={progress} max="100" style:width="100%" />
```

I'm suggesting instead using something like: https://cssloaders.github.io which uses css animations. I'd guess this would not be subject to the same "freezing" so long as the animation is *started* before the main thread gets blocked.

michaelkirk:
(I might be wrong though)

PR #357, opened 2025-05-08
If a user loads an empty project or deletes all neighbourhoods, don't

dabreegster:
show them an empty pick screen

We've had a few bits of confused feedback from people loading an empty project and not understanding what to do (despite the primary "add neighbourhood" button. So just forcibly prevent people from winding up on this screen without anything defined.


michaelkirk (approved):
Great! I had the same thought, but just never got around to it.

PR #359, opened 2025-05-08
Handle edge cases with no main roads

dabreegster:
It's totally possible to manually draw an area that has no main roads at all. Previously there was a popup error, but with this PR, it's now possible, but the result is pretty boring, because there are no shortcuts.
image

The real use of this is with a new button to erase all main roads at once. If you're working on a circulation plan in a huge area like this, using the eraser tool or manually clicking all of the main roads on the inside is really, really tedious:

image

It's easier in this case to just blank out all the main roads, then draw only the main roads around the outside, and any on the inside that should retain through-traffic. We can call out this special use case in the user guide.

### DIFF
--- a/backend/src/lib.rs
+++ b/backend/src/lib.rs
@@ -400,6 +400,14 @@ impl LTN {
         self.after_main_road_edit()
     }
 
+    #[wasm_bindgen(js_name = eraseAllMainRoads)]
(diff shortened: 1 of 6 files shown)


michaelkirk (approved):
LGTM

PR #362, opened 2025-05-12
Reorganize CNT asset storage and gzip most files

### LINKED ISSUE #239: Serve gzipped files
The osm.pbf and bincoded demand files and prioritization inputs are all stored on Cloudflare R2 and served from there. They're not stored or sent compressed, but they should be, for less to download. I've had some trouble in the past uploading a .gz file and getting the `Content-Encoding: gzip` header to come from Cloudflare. I'll figure this out before launch.

dabreegster:
FIXES #239 and #255

This PR reorganizes the directory structure for CNT assets, from the repetitive "cnt_demand/demand_LAD_foo.bin" structure to "cnt/demand/LAD_foo.bin.gz". Nesting all the CNT directories in one "cnt/" directory lets me organize my local cache of cloudflare R2 locally and have a script to sync it properly.

And secondly, this PR gzips most assets, achieving a big reduction in download time. Browsers will magically gunzip before handing to WASM, thanks to the `content-encoding: gzip` header. For future reference, to get this served by cloudflare (objects stored in R2, with a custom domain `assets.cnt.scot` wired up), it appears impossible to use `rclone`, despite vague docs online suggesting otherwise. After uploading the gzipped objects, I'm doing something like this to set the headers:

```
aws s3 --profile cloudflare \
        --endpoint-url https://FOO.r2.cloudflarestorage.com \
        cp s3://BAR/ s3://BAR/ \
        --exclude '*' \
        --include '*.bin.gz' \
        --include '*.pbf.gz' \
        --no-guess-mime-type \
        --content-type='application/octet-stream' \
        --content-encoding='gzip' \
        --metadata-directive='REPLACE' \
        --recursive
```

@michaelkirk, you'll need to rerun `bin/download-local-dev-data.sh` or regenerate things locally.

### DIFF
--- a/backend/Cargo.toml
+++ b/backend/Cargo.toml
@@ -13,6 +13,7 @@ console_error_panic_hook = "0.1.6"
 console_log = "1.0.0"
 contour = "0.12.0"
 fast_paths = "1.0.0"
+flate2 = "1.1.1"
 geo.workspace = true
 geojson.workspace = true
 itertools = "0.14.0"
@@ -35,9 +36,8 @@ web-sys = { version = "0.3.64", features = ["console"] }
 web-time = "1.0.0"
 
 [dev-dependencies]
-criterion = "0.5.1"
 approx = "0.5.1"
-
+criterion = "0.5.1"
 
 [[bench]]
 name = "boundary_stats"
--- a/data_prep/.gitignore
+++ b/data_prep/.gitignore
@@ -1,3 +1,0 @@
-od.csv
-zones.geojson
-out/
--- a/data_prep/scotland/README.md
+++ b/data_prep/scotland/README.md
@@ -12,7 +12,7 @@ unzip SG_IntermediateZoneBdry_2011.zip
 ogr2ogr zones.geojson -t_srs EPSG:4326 -nlt PROMOTE_TO_MULTI SG_IntermediateZone_Bdry_2011.shp -sql 'SELECT InterZone as name FROM SG_IntermediateZone_Bdry_2011'
 rm -f SG_IntermediateZone_Bdry_2011* SG_IntermediateZoneBdry_2011.zip
 
-mkdir -p demand
+mkdir -p ../../web/public/cnt/demand ../../web/public/cnt/prioritization
 cargo run --release --bin generate_od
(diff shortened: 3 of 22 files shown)


backend/Cargo.toml
```diff
@@ -13,6 +13,7 @@ console_error_panic_hook = "0.1.6"
 console_log = "1.0.0"
 contour = "0.12.0"
 fast_paths = "1.0.0"
+flate2 = "1.1.1"
```
dabreegster, line 16: Temporary hack, this should be dev-dependencies; it's only for the benchmark code. I think we need to follow the note in `lib.rs`:

```
// TODO: We could hide this behind a feature flag - it's used by both tests and benches
pub mod test_fixtures;
```

I couldn't figure out how to feature-guard for "in test/benchmark mode". I'll wire this up in a separate followup

data_prep/.gitignore
dabreegster, line 1: Some entries here became stale, I merged with the one repo-wide file to maintain more easily

data_prep/scotland/README.md
```diff
@@ -12,7 +12,7 @@ unzip SG_IntermediateZoneBdry_2011.zip
 ogr2ogr zones.geojson -t_srs EPSG:4326 -nlt PROMOTE_TO_MULTI SG_IntermediateZone_Bdry_2011.shp -sql 'SELECT InterZone as name FROM SG_IntermediateZone_Bdry_2011'
 rm -f SG_IntermediateZone_Bdry_2011* SG_IntermediateZoneBdry_2011.zip
 
-mkdir -p demand
+mkdir -p ../../web/public/cnt/demand ../../web/public/cnt/prioritization
```
dabreegster, line 15: In all but one script, I've made the data prep scripts directly write to `web/public/cnt`, instead of have temporary output directories and then moving things over. Before running the scripts, you can move / symlink / blank out your `web/public/cnt` as desired

dabreegster (commented):
I've tested locally with and without my local `public/` cache and made sure benchmarks run, so I'll merge this and deploy to prod too, so I can clean up some of my old cloudflare mess. Async feedback welcome.

PR #363, opened 2025-05-13
Set up a special deployment for England

dabreegster:
image

Most existing LTN users are in England, so I'm going to make things nicer like CNT/Scotland here too. This PR just sets up the boundaries (based on 2024 LADs, which're a bit small in places, but a reasonable start). I'll follow up later with demand data, contextual layers, and prioritization data (much of which is very similar to Scotland anyway).

The screenshot above shows two other related changes:

1) Going back to an old conversation around how users will identify the version of the LTN tool and switch, I've added a widget to switch between global/England/Scotland on the first page. I'm **not** showing it when the user is on CNT, since that's a special branded deployment, and I don't think users there want to be directed away. (But I'll discuss and maybe add it later.)

2) Instead of listing the boundary names in England and Scotland as a three column alphabetized list, use a dropdown. This is a design decision we made in NPW, and I think it makes sense in general. Quickly spotting a name in 3 columns is hard, and it's a very long list.

I'll upload the new data in a few hours, then merge. If you're reviewing by commit, note they're a little messy -- I fixed problems with an earlier commit in a later commit.


data_prep/england/build_all_areas.sh
dabreegster, line 1: This is a copy of the Scotland one with some things changed. Short enough to not muck around with trying to generalize it.

data_prep/geojson_to_osmium_extracts.py
dabreegster, line 1: I did "refactor" this script to work for both uses

web/package.json
```diff
@@ -35,6 +35,7 @@
     "@picocss/pico": "^2.0.6",
     "@turf/along": "^7.1",
     "@turf/distance": "^7.2.0",
+    "flag-icons": "^7.3.2",
```
dabreegster, line 38: 28KB minified CSS, not sure if magic tree-shaking happens or not, but it's not big

web/src/title/AppFocusSwitcher.svelte
```diff
@@ -0,0 +1,41 @@
...
+  import { Globe } from "lucide-svelte";
+  import { onMount } from "svelte";
+  import { HelpButton, initTooltips } from "../common";
+  import "flag-icons/css/flag-icons.min.css";
+  import { appFocus } from "../stores";
+
+  onMount(initTooltips);
+</script>
+
+<span style="display: flex; flex-direction: row; height: 80%; gap: 4px">
```
dabreegster, line 11: image
Alignment / height isn't perfect, but good enough start hopefully

dabreegster:
@mvl22, in case you're interested, https://a-b-street.github.io/ltn/england.html is now easily usable in Cambridge and elsewhere

michaelkirk (commented):
Seems reasonable! (Though that layout could use some alignment love).

michaelkirk:
Another nit: there's now an undesirable horizontal scroll on the landing page.

michaelkirk:
Sorry, I should have been more clear. I think it's visible in your screenshot:

image

The new dropdown is beyond the right edge. It seems to occur at any width as far as I can tell.

dabreegster:
Ohh, it's the select, not the top bar! Thanks, I'll fix tonight or tomorrow. Not sure why I didn't notice this.

dabreegster:
84bc46f5b6e08880385b2bbb61cfb04818ff833e fixes both -- thanks for pressing me on this. Slowly CSS is making sense (now that pico defaults don't constantly get in the way)

michaelkirk:
Looks much better, thank you!

PR #373, opened 2025-06-02
Preprocess dog-leg intersections

### LINKED ISSUE #371: Diagonal filters on dog-leg intersections
Diagonal filters only work on 4-way intersections. But some cases are very slightly offset / "dog-legged":

https://www.openstreetmap.org/#map=19/55.959022/-3.134598

Image

Users expect these to act like four-way intersections, because in practice they're real candidates for a diagonal filter.

## Idea 1

Can we detect these very short edges and allow diagonal filters to be placed there? This would complicate the data model and how we apply them.

## Idea 2

Related to #350, can we transform the graph upfront? This feels like the cleaner solution, but I'm not sure how to represent the geometry. I think two of the roads will have to share the short segment. Need to revisit the movement logic, clockwise sorting, etc to do this correctly.

dabreegster:
FIXES #371.

A user brought up some cases where they unexpectedly couldn't create a diagonal filter, because a 4-way intersection is slightly offset -- a "dog leg":
image
This PR preprocesses the OSM network, detecting these cases and turning into a regular 4-way intersection, so then the diagonal filter 
image

I worked through this approach over in https://github.com/dabreegster/road-bundler/blob/main/backend/src/dog_leg.rs, a new experiment for preprocessing OSM for many purposes, and then adapted the solution here.


backend/src/osm_tests/dog_legs.osm.xml
dabreegster, line 1: Reminder you can use https://dabreegster.github.io/geojson-viewer to quickly view this, and now also https://dabreegster.github.io/road-bundler
image
The left case is a real dog-leg, and the right is a case that should not be detected and transformed

backend/src/create/dog_leg.rs
```diff
@@ -0,0 +1,157 @@
...
+    if Euclidean.length(&edge.linestring) > 5.0 {
+        return None;
+    }
+    if !edge.osm_tags.has("name") {
+        return None;
+    }
+    let src_i = &graph.intersections[&edge.src];
+    let dst_i = &graph.intersections[&edge.dst];
+    let mut src_edges = src_i.edges.clone();
+    let mut dst_edges = dst_i.edges.clone();
```
michaelkirk, line 40: It might not be a big deal, but it think we could avoid these clones if this method is expensive.

backend/src/create/dog_leg.rs
```diff
@@ -0,0 +1,157 @@
...
+    graph.intersections.insert(
+        new_intersection,
+        Intersection {
+            id: new_intersection,
+            edges: Vec::new(),
+            point: midpt,
+            // OSM IDs are only used for debugging. Rather than change the representation here to
+            // allow for "synthetic" intersections without an OSM node, just borrow one of the
+            // endpoints.
+            osm_node: graph.intersections[&src].osm_node,
```
michaelkirk, line 92: I understand how this currently isn't causing any problems, but it might be confusing at some point. One thing I sometimes see with synthetic features is using a negative id as a sentinel value. It doesn't seem like a big deal though.

michaelkirk (commented):
I can imagine the new functionality is more in line with what users expect, and the implementation seems reasonable 👍

PR #375, opened 2025-06-03
User guide

### LINKED ISSUE #45: Docs
- [ ] Fill out the project README
- [ ] Developer / contributor guide
- [ ] Fix up all the in-app help text

dabreegster:
FIXES #45 

This turns an internal google doc into a real user guide linked  from the app. It can be accessed from the first screen or at any time in the top-right corner:
image

The guide opens in a new tab. The styling there is simple, using svelte-toc for a floating right-hand table of contents:
image


web/src/App.svelte
```diff
@@ -157,10 +166,17 @@
       class="pico"
       style="display: flex; align-items: center; gap: 8px;"
     >
-      <button class="outline" on:click={() => ($showAbout = true)}>
+      <button class="outline" on:click={() => (showAbout = true)}>
         <img src={logo} style="height: 32px;" alt="A/B Street logo" />
       </button>
       <span bind:this={topDiv} style="width: 100%" />
+      <button
```
dabreegster, line 173: It seems useful to always have a link to the manual handy. Hopefully the help icon is obvious enough. I also tried this one:
image

web/src/PickNeighbourhoodMode.svelte
```diff
@@ -170,19 +169,6 @@
         </li>
         <li>
           {pageTitle($mode.mode)}
-          <HelpButton>
```
dabreegster, line 173: There were just two modes with these. I got rid of it for now. Maybe in the future, we could try putting the help button over here again, and making it open to the appropriate section of the manual

web/src/user_guide/Page.svelte
```diff
@@ -0,0 +1,518 @@
+<script lang="ts">
+  import "@picocss/pico/css/pico.conditional.jade.min.css";
+  import Toc from "svelte-toc";
+  import About from "../About.svelte";
+  import image01 from "./image01.webp?url";
```
dabreegster, line 5: Vite is kind of annoying here, I don't know how to avoid this boilerplate. When I've used the static assets directory in the past, I've hit every combination of headaches making it work on GH Pages vs locally vs on a domain without a leading base path.

web/src/user_guide/Page.svelte
```diff
@@ -0,0 +1,518 @@
...
+  import image29 from "./image29.webp?url";
+  import image30 from "./image30.webp?url";
+  import image31 from "./image31.webp?url";
+</script>
+
+<div class="container">
+  <div class="contents">
+    <h1>User guide for the Connected Neighbourhood Tool</h1>
+
+    <h2 id="overview">Overview</h2>
```
dabreegster, line 42: Converting this from google docs was a slight chore. I converted to HTML to get the images, but the HTML it generates is so messy. So I converted to markdown, then did `npx showdown makehtml -c ghCompatibleHeaderId -i CNT\ user\ guide.md -o simple.html`. It tried to inline images in paragraphs with linebreaks, and also got confused about ordered lists. I fixed a bunch of stuff manually.

web/src/user_guide/Page.svelte
```diff
@@ -0,0 +1,518 @@
...
+  import About from "../About.svelte";
+  import image01 from "./image01.webp?url";
+  import image02 from "./image02.webp?url";
+  import image03 from "./image03.webp?url";
+  import image04 from "./image04.webp?url";
+  import image05 from "./image05.webp?url";
+  import image06 from "./image06.webp?url";
+  import image07 from "./image07.webp?url";
+  import image08 from "./image08.webp?url";
+  import image09 from "./image09.webp?url";
```
dabreegster, line 13: The original images were pngs, up to 2MB each. Webp is tiny and looks identical, and https://caniuse.com/webp says it's stable since 2020.

web/src/user_guide/Page.svelte
```diff
@@ -0,0 +1,518 @@
...
+
+    <h2 id="adding-a-neighbourhood">Adding a neighbourhood</h2>
+    <p>
+      Start using the tool by going to <a href="https://cnt.scot/"
+        >https://cnt.scot/</a
+      >
+      and picking your study area on the map or from the list. The study areas are
+      defined by Local Authority Districts.
+    </p>
+    <img alt="" src={image01} />
```
dabreegster, line 72: No alt text anywhere :\  There's 31 images to possibly caption, and the prose surrounding the image makes them clear in many cases. Could write captions in the future.

web/src/user_guide/Page.svelte
```diff
@@ -0,0 +1,518 @@
...
+  import image27 from "./image27.webp?url";
+  import image28 from "./image28.webp?url";
+  import image29 from "./image29.webp?url";
+  import image30 from "./image30.webp?url";
+  import image31 from "./image31.webp?url";
+</script>
+
+<div class="container">
+  <div class="contents">
+    <h1>User guide for the Connected Neighbourhood Tool</h1>
```
dabreegster, line 40: I thought about including a link back to the tool, but it'd be a small bit of plumbing to remember if the user was on global/England/CNT. Since it opens in a new tab, hopefully people know how to switch back.

dabreegster:
I'll merge this now to get final feedback more easily, but async thoughts really welcome. Another vague question is about the image sizing -- sometimes they might be too big? The source images are sometimes the full-screen, but sometimes just a clip. I'm not sure what standard practice is hre.

web/src/user_guide/Page.svelte
michaelkirk: If you ever do need to use png, I've seen some moderate improvements by using something like pngquant. 

Though I assume webp compresses even better.

web/src/user_guide/Page.svelte
michaelkirk: Maybe better to remove the empty alt tag? I'm not sure if that might confuse screen readers.

michaelkirk:
Something seems out of whack with the images when I go to: https://a-b-street.github.io/ltn/user_guide.html

At least the first 10 or so images don't  seem to correspond to the text.

e.g.

## first image

Screenshot 2025-06-06 at 17 37 59

I'd expect a screenshot of the LAD picker not the neighbourhood boundary picker.

## second and third images

Screenshot 2025-06-06 at 17 39 07

Here I'd expect to see the neighbourhood boundary picker, but it looks like something showing shortcuts and modal filter placement. 

Really I can't tell if *any* of the photos are in the right place. Let me know if you need more details.

PR #379, opened 2025-06-10
Show a neighbourhood before edits

### LINKED ISSUE #376: Add toggle to quickly show the "before" of a LTN
When doing a presentation it would be useful to be able to quickly show the current state, before any restrictions.

I tried to create a new neighborhood with the same limits in the same project, but the restrictions are already applied to it.

I guess creating a different project would work, but it's not convenient to switch from one to the other.

dabreegster:
Fixes most of #376.

This introduces a new control to revert all edits and just display the original state of a neighbourhood, so the user can quickly compare.
[Screencast from 13-06-25 08:38:01.webm](https://github.com/user-attachments/assets/912869e8-6465-4ff3-9110-b84e1643e4e1)

Try it out: https://a-b-street.github.io/ltn/before_after_V2
@XioNoX and @TFCx, please see if this achieves what you're trying to do

Questions:

- Is the "Draw roads: edits (either filter or direction)" still useful after this? I think maybe so, if it's a complex area
  - From discussion: YES, keep it
- Does it make sense to also show the original main road classification? Do you ever want to see shortcuts on a current main road, but without any new filters?
  - From discussion: It could be useful sometimes to see the new classification, but the reasonable choice for now is to show the original classification
- Is it OK to only have this feature in the main neighbourhood editing screen? Route mode, predict impact, and impact to one destination all already have a comparison of before/after. Is it useful to use the detailed shortcuts tool on the original state? Or just see all original filters on the pick neighbourhood screen?
  - From discussion: It could help in pick neighbourhood, but the other places don't need it. For now, just have in the edit neighbourhood view.

- [x] Show original turn restrictions
- [ ] Don't show the pointer cursor
- [x] Figure out what happens if you enter this mode in the middle of drawing a turn restriction or something complex
- [x] Figure out what happens if you exit this mode suddenly
- [x] Support animating shortcuts
- [x] Cache the original GJs

Implementation notes:

- I first tried a stateful approach in the backend, making a call to actually revert all of the commands and bump mutationCounter as usual. But restoring the state gets confusing, and so many things need to change on the frontend anyway -- no autosaving while showing the original state. No edits allowed, so all the layer interactivity changes anyway.

### DIFF
--- a/web/src/edit/NeighbourhoodMode.svelte
+++ b/web/src/edit/NeighbourhoodMode.svelte
@@ -1,6 +1,14 @@
 <script lang="ts">
   import type { Feature, FeatureCollection, LineString } from "geojson";
-  import { Eraser, Pointer, Redo, Route, Trash2, Undo, Paintbrush } from "lucide-svelte";
+  import {
+    Eraser,
+    Paintbrush,
+    Pointer,
+    Redo,
+    Route,
+    Trash2,
+    Undo,
+  } from "lucide-svelte";
   import type { LngLat, MapMouseEvent } from "maplibre-gl";
   import { onDestroy, onMount } from "svelte";
   import {
@@ -31,6 +39,7 @@
     roadLineWidth,
   } from "../common";
   import { speedColorScale, speedLimits } from "../common/colors";
+  import type { Waypoint } from "../common/draw_area/stores";
   import type { Intersection } from "../common/Intersection";
   import { ModalFilterType } from "../common/ModalFilterType";
   import NeighbourhoodBoundarySummary from "../common/NeighbourhoodBoundarySummary.svelte";
@@ -54,6 +63,7 @@
     mutationCounter,
     roadStyle,
     saveCurrentProject,
+    showBeforeEdits,
     thickRoadsForShortcuts,
   } from "../stores";
   import type {
@@ -63,6 +73,7 @@
   } from "../wasm";
   import ChangeFilterModal from "./ChangeFilterModal.svelte";
   import FreehandLine from "./FreehandLine.svelte";
+  import ShowBeforeEdits from "./ShowBeforeEdits.svelte";
   import SnapRouteSelector from "./SnapRouteSelector.svelte";
 
   // Caller is responsible for doing backend.setCurrentNeighbourhood
@@ -82,6 +93,8 @@
     | {
         kind: "main-roads";
         tool: "toggle" | "snap-route-main" | "snap-route-erase";
+        // Only used for snap-route-main or snap-route-erase
+        waypoints: Waypoint[];
       };
   function startTurnRestrictionAction(): Action {
     return {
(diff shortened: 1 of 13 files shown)


TFCx:
For a first version, it does what I expect.

> Is the "Draw roads: edits (either filter or direction)" still useful after this? I think maybe so, if it's a complex area

Currently, edits do not catch the eyes sufficently. The "Draw roads: edits" allow the edits to be more visibles... But there's probably a solution where both tools can be merged.

> Does it make sense to also show the original main road classification? Do you ever want to see shortcuts on a current main road, but without any new filters?

For me, the current implementation is correct: I want to see which roads were declassified.

> Is it OK to only have this feature in the main neighbourhood editing screen?

The use of a checkbox + its position in the layout seems a bit strange for me.

I was wondering if it would be easy to split the map layout into two when visualizing before/after ? And bound the coordinates on both maps ? A bit like but with the two halves centered on the same point
image

dabreegster:
Thanks for the quick feedback!

> The use of a checkbox + its position in the layout seems a bit strange for me.

I agree... I tried on the top-right corner of the map first, or hidden inside the layers control, but it felt less prominent and discoverable. And the effect of disabling the editing tools felt odd, because it was on the other side of the screen. Maybe if the editing tools / show before edits were styled like two tabs?

> I was wondering if it would be easy to split the map layout into two when visualizing before/after ? And bound the coordinates on both maps ? A bit like but with the two halves centered on the same point

Hmm, this does seem useful, but I think it's harder to do. I'll experiment with this, but ideally some improvements on this current version can go in as a first version.

XioNoX:
Thanks!! That PR solves exactly my usecase.

> Is the "Draw roads: edits (either filter or direction)" still useful after this? I think maybe so, if it's a complex area

Yup, still useful as mentioned by Jean-David.

> Does it make sense to also show the original main road classification? Do you ever want to see shortcuts on a current main road, but without any new filters?

I think both usecase are valid. For example on one of the neighborhood I work on, there is a short living street section (https://www.openstreetmap.org/way/462948349) which is effectively a regular street (only some signs, cars drive fast).
So I re-classified it at the very beginning, and when doing a before/after I'd prefer to still see it as main road.
But as you said, "perfect is the enemy of good" and I don't have a strong opinion for now (nor an idea on how to make it better without a 1000 knobs). 

> Is it OK to only have this feature in the main neighbourhood editing screen? 

Yeah that looks good to me. Maybe a toggle instead of a checkbox, but that's just UI :)

>  Is it useful to use the detailed shortcuts tool on the original state?

I haven't had the need for that.

> Or just see all original filters on the pick neighbourhood screen?

I think being able to see the shortcuts on the "initial state" is important, the way it's showed on the "main neighbourhood editing screen" looks good to me.

web/src/edit/NeighbourhoodMode.svelte
```diff
@@ -374,213 +401,238 @@
...
   </div>
   <div slot="sidebar">
-    <h2>Editing tools</h2>
     <div
-      class="tool-palette"
-      style="display: flex; justify-content: space-between; flex-wrap: wrap; gap: 6px;"
+      style="display: flex; justify-content: space-between; align-items: center"
     >
+      <h2>Editing tools</h2>
+      <label>
```
dabreegster, line 408: Happy for other ideas about how to style / place the toggle and to show the editing controls are disabled
Screenshot from 2025-06-13 08-44-04

dabreegster:
Merging now ahead of a conference with demos, but happy to address feedback later. #376 has two tasks for followup also. Thanks @TFCx and @XioNoX for all the ideas and feedback!

PR #381, opened 2025-06-10
Include more shortcuts, by penalizing main roads

### LINKED ISSUE #377: Shortcut calculation can miss cases with higher speed main roads
Around `#14.41/55.95014/-3.14529` in Edinburgh, Willowbrae Road / the A1 is 30mph, and the residential streets nearby are 20mph. In the big yellow cell, there are no shortcuts, because the fastest route between all of the border intersections (on the A1) are using the A1, not cutting through:

Image

This feels intuitively wrong. The router used in `shortcuts.rs` sets a `main_road_penalty` of 1 (no penalty). If I set this to 2 (main roads are twice as slow due to congestion, making their effective speed 15mph, slower than the residential roads), then I get a shortcuts pattern that I expected:

Image

Should we make this change? Some relevant parts from https://cnt.scot/user_guide#understanding-the-map-shortcuts:

"Shortcuts show the possible routes through a neighbourhood a driver might take **when avoiding main roads**."

"For every combination, calculate the **shortest** driving route, restricted to not cross a main road"

If we decide to not make this change, at the very least this requires clarification in the manual. I'm inclined to use a penalty of 2 to encourage more shortcuts in cases like this.

dabreegster:
FIXES #377 

Demo: https://a-b-street.github.io/ltn/shortcut_calculation

- [x] Update user guide
- [x] Update changelog

I need to examine the test diffs and understand how it affects more cases. Anybody following along, please feel to try and see if your intuition about shortcuts near faster main roads shows this version is better or not


dabreegster:
#377 has an example with a big difference in Edinburgh. The new version feels correct to me.

Bristol east changes from:
image
to:
image
The before image looks totally wrong to me -- it implies only that long east/west Beaufort Road has a possible problem, and the other streets are effectively cul-de-sacs. That's definitely wrong. An example shortcut from the new version cuts the corner and avoids a traffic signal. The new version feels much more correct to me.
image

Bristol west has only a small difference. Before:
image
After:
image
There's some shortcuts picked up on the E/W road above Hayward. (I really need to get Ubuntu screenshot annotations working again...)

Finally, Strasbourg has a small change too. One road with new shortcuts picked up:
image

All of these cases look more intuitive to me in the after case. Even though the shortcut isn't the **fastest** way through if the main roads have no delays, the point is that people try shortcuts when the main roads are slower.

Issue #24, opened 2024-03-02
Polish projects UX


dabreegster:
- [ ] Include a last modified time in projects and show this on the title screen
- [x] Save as / make copy of a project?
- [x] Title screen shouldn't have a map at all
- [x] Internally use `ltn_name.geojson`, but don't show the prefix and suffix in the UI
- [ ] Warn when clobbering an existing name for both projects and neighbourhoods
- [x] More granular progress bar
- [ ] Handle multiple proposals in one project?

PR #394, opened 2025-06-24
Copy a project and quickly switch

### LINKED ISSUE #307: Easily copy a project
A user wants to easily compare:

- the baseline / existing situation somewhere
- a main design option
- some alternative design options/variants

We decided not to make the data model and UI more complicated to support this natively. (Maybe "yet" -- in a cloud/collaborative service future, we could revisit.) Instead, the recommendation is just to use one project file to represent each of these cases, and use multiple browser tabs to switch. Concretely, this means we need to make it easy for users to make a copy of a project file and give it a new name.

dabreegster:
[Screencast from 24-06-25 13:41:51.webm](https://github.com/user-attachments/assets/2a92964f-7ee6-411b-b29a-cca5f5cfd990)
FIXES #307, helps #24, and was discussed in #376. CC @XioNoX and @TFCx

This adds:

1) A button to copy a project on the very first screen
2) A button to copy a project (and immediately switch to it) from the pick neighbourhood screen
3) A dropdown to quickly switch between projects in the same study area, on the pick neighbourhood screen

This makes it much more fluid to manage multiple "proposals" for the same study area. You could define a neighbourhood boundary, copy the project, and switch between the two, making different edits. It requires always going back to the pick neighbourhood screen to switch, but this is much faster than the previous flow.


web/src/pick_neighbourhood/PickNeighbourhoodMode.svelte
```diff
@@ -172,7 +138,9 @@
           <ModeLink mode={{ mode: "title" }} />
         </li>
         <li>
-          {pageTitle($mode.mode)}
+          {#key $currentProjectID}
```
dabreegster, line 141: We need to update this when project name changes

web/src/title/TitleMode.svelte
```diff
@@ -204,4 +226,8 @@
   .project-list li {
     padding-left: 1em;
   }
+
+  .project-name {
+    overflow-wrap: anywhere;
```
dabreegster, line 231: image
The wrapping looks reasonable to me, keeps the buttons always the same size/location

dabreegster (commented):
Oh yeah, I also will update the manual before merging this PR, but it's otherwise ready for review

Issue #404, opened 2025-08-16
Consider using Bootstrap


dabreegster:
In https://github.com/a-b-street/svelte-utils/issues/4, I've started consolidating Svelte helpers in a few other projects. There are two particular ideas to apply here:

1) Switching to Svelte 5, to unblock #368. From my experience so far with 5, I'm not yet convinced going all-in on the new runes is worth it. But we should be able to keep svelte 4 style, but still run with 5 and use wuchale, I think.

2) I've switched from picocss to Bootstrap in a few projects, and it's... less bad than pico. At least more maintained and complete. https://a-b-street.github.io/severance_snape/ and https://dabreegster.github.io/road-bundler/ are two projects converted to using this. Is it worth trying to do the same here? @michaelkirk heroically worked around dozens of pico issues here, especially maplibre/pico interactions. I don't want to go through that kind of effort again just to fit in well with Bootstrap, since Bootstrap is... fine, but not incredible.

Both of these points relate to `svelte-utils` too. We can't upgrade usage of that in this repo until we do both of the above. I've made some (I think) nicer basemap pickers and right-click controls in those newer projects, but I haven't thought about if they'd fit in well here yet.

PR #405, opened 2025-08-16
Switch to vite-plugin-wasm and update vite

dabreegster:
This is a prereq to attempting a Svelte 5 upgrade (for #404). The old vite-plugin-wasm-pack plugin has been partly broken from when I first started using it a few years ago; the latest published version has never worked. vite-plugin-wasm is the newer thing compatible with modern Vite.

### DIFF
--- a/.github/workflows/web.yml
+++ b/.github/workflows/web.yml
@@ -44,10 +44,6 @@ jobs:
               npm run build --if-present -- --base=/ltn/$branch
             fi
 
-            # TODO Vite suddenly broke and I have no idea why. Hack around it.
-            # new URL('backend_bg.wasm', import.meta.url) doesn't exist at build time, it will remain unchanged to be resolved at runtime
-            ln -s assets/backend_bg.wasm dist/
-
             if [ "$branch" == "main" ]; then
               mv dist/* deployme
             else
--- a/web/package.json
+++ b/web/package.json
@@ -16,7 +16,7 @@
   },
   "devDependencies": {
     "@ianvs/prettier-plugin-sort-imports": "^4.6.2",
-    "@sveltejs/vite-plugin-svelte": "^2.5.3",
+    "@sveltejs/vite-plugin-svelte": "^3.1.0",
     "@tsconfig/svelte": "^5.0.4",
     "@types/geojson": "^7946.0.16",
     "@types/node": "^24.3.0",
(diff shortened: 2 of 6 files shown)


.github/workflows/web.yml
```diff
@@ -44,10 +44,6 @@ jobs:
               npm run build --if-present -- --base=/ltn/$branch
             fi
 
-            # TODO Vite suddenly broke and I have no idea why. Hack around it.
```
dabreegster, line 47: This was some bug with the old plugin

web/package.json
```diff
@@ -41,7 +41,10 @@
     "route-snapper-ts": "^0.0.8",
     "svelte-maplibre": "^0.9.14",
     "svelte-toc": "^0.5.9",
-    "svelte-utils": "github:a-b-street/svelte-utils",
+    "svelte-utils": "github:a-b-street/svelte-utils#6a5b755ae7f8216c5b8996333b16221f1b344595",
```
dabreegster, line 44: We need to pin a few things, because in the latest commit in these packages, I've switched to Svelte 5

web/vite.config.ts
```diff
@@ -18,11 +18,20 @@ export default defineConfig({
   plugins: [
     // disable hot module reloading during test runs
     svelte({ hot: !process.env.VITEST }),
-    wasmPack(["../backend"], ["route-snapper"])
+    wasm(),
   ],
+  server: {
+    fs: {
+      allow: ["./", "../backend/pkg"],
```
dabreegster, line 25: I dunno why importing from "backend" doesn't work, but it doesn't, and we need to tell vite it's OK to look out of the web/ dir

web/vite.config.ts
```diff
@@ -18,11 +18,20 @@ export default defineConfig({
...
   test: {
-    environment: 'jsdom',
-    setupFiles: ['./vitest.setup.ts'],
-  }
-})
+    environment: "jsdom",
+    setupFiles: ["./vitest.setup.ts"],
+  },
+  optimizeDeps: {
+    // Something breaks WASM import
```
dabreegster, line 34: Not sure what's going on here either, but this works locally

dabreegster (commented):
Works locally, it's possible something about the build / GH actions will break, but I'll... merge and see!

PR #406, opened 2025-08-17
Switch to svelte 5

dabreegster:
For #404. Updates to Svelte 5 in the minimal way, not yet using runes anywhere.

Almost ready, but a long tail of small issues...

### DIFF
--- a/web/cnt.html
+++ b/web/cnt.html
@@ -8,9 +8,10 @@
   <body>
     <div id="app"></div>
     <script type="module">
+      import { mount } from "svelte";
       import App from "./src/App.svelte";
 
-      new App({
+      mount(App, {
         target: document.getElementById("app"),
         props: {
           appFocus: "cnt",
--- a/web/england.html
+++ b/web/england.html
@@ -8,9 +8,10 @@
   <body>
     <div id="app"></div>
     <script type="module">
+      import { mount } from "svelte";
       import App from "./src/App.svelte";
 
-      new App({
+      mount(App, {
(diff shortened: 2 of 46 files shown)


dabreegster:
I switched to runes and fixed the above bug, but found another -- z-ordering of the `intersection-filters` diagonal filters layer is wrong. We make a new layer through svelte-maplibre, it does the correct `beforeId`, and then we set up another layer. It does `getLayer` on the first, but it's not always present yet, so the z-ordering gets messed up.

We could...
- always wait for `styledata` events like https://github.com/MIERUNE/svelte-maplibre-gl does
- after adding all the layers, try doing `moveLayer` everywhere
- something else?

I'm more and more nervous about how enormous this change is becoming and how quiet some of the bugs are.

web/src/edit/FreehandLine.svelte
```diff
@@ -52,17 +53,14 @@
 
       line.geometry.coordinates.push(next);
       line = line;
-      if (line.geometry.coordinates.length % 10 == 0) {
-        dispatch("progress", line);
```
dabreegster, line 56: progress event was unused

dabreegster (commented):
I've gone through all the code changes and they seem fine. Going to do a last round of testing, then try and merge this.

dabreegster:
I'm calling this good enough and merging. #407 tracks one small behavioral fallout (freehand line drawing) and some code smells

dabreegster:
Rolled back, track #407 for updates

PR #408, opened 2025-08-30
Set up i18n using wuchale

dabreegster:
This PR wires up https://wuchale.dev/ and configures French and Hungarian. The proof it's working is "Read the user guide" translated on the splash page -- please do fix that translation if Google was wrong! Switching language just reloads the entire page, because seemingly swapping everything out dynamically doesn't work too well (or requires the entire App inside the `await loadLocale` to cleanly unmount and re-remount, which... does not work at all).

The language picker is shown on the splash page:
image
And is always in the top-right layers/settings menu:
image
Suggestions welcome for different designs here.

Please give it a try and let me know if you have any issues, @Kandico, @TFCx, @rouelibre1. There could be some strings that aren't detected yet (HTML titles? The `ModeLink`s in the breadcrumbs?). We'll sort these out as y'all discover them.


K1DV5:
Hello @dabreegster, a couple of suggestions:

- You don't actually need to do `npx wuchale init` every time you add a new string as that would be annoying. Unless you disable it, it should do its job in the background. The HMR integration is something I've spent a lot of time on to avoid doing anything extra.
- I see that some of the messages are not translated yet, and so you may also like that you can [use Gemini to translate everything](https://wuchale.dev/guides/gemini/) and this works even live during dev.
- You may find the tip on [getting the language names automatically](https://wuchale.dev/guides/tips/#language-names) interesting.

Kandico:
> Hello @dabreegster, a couple of suggestions:
> 
> * You don't actually need to do `npx wuchale init` every time you add a new string as that would be annoying. Unless you disable it, it should do its job in the background. The HMR integration is something I've spent a lot of time on to avoid doing anything extra.
> * I see that some of the messages are not translated yet, and so you may also like that you can [use Gemini to translate everything](https://wuchale.dev/guides/gemini/) and this works even live during dev.
> * You may find the tip on [getting the language names automatically](https://wuchale.dev/guides/tips/#language-names) interesting.

I tried and ai translation, regardless of the llm didn't work great in Hungarian when it comes to ab/street. I will be able to reply in greater detail and translate most of the app next week.

PR #410, opened 2025-09-13
Start French translation

XioNoX:
First time doing that. Probably best by doing small iterations, especially depending on how the string lengths differences impact the UI.

### DIFF
--- a/web/src/locales/fr.po
+++ b/web/src/locales/fr.po
@@ -17,16 +17,20 @@ msgid ""
 "This is the second version of the A/B Street LTN tool, <0>an open source project</0>. This version, called the Connected Neighbourhoods Tool, has been funded by <1>Transport Scotland</1> and designed by <2>Sustrans Scotland</2>.\n"
 "The team includes:"
 msgstr ""
+"Ceci est la seconde version de l'outil A/B Street LTN, <0>un projet open source</0>. Cette version, nommée Connected Neighbourhoods Tool, a été financé par <1>Transport Scotland</1> et conçue par <2>Sustrans Scotland</2>.\n"
+"L'équipe est constituée de :"
 
 #: src/About.svelte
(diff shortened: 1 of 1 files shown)


web/src/locales/fr.po
```diff
@@ -1846,201 +1852,217 @@ msgstr ""
...
 #: src/edit/NeighbourhoodMode.svelte
 msgid "Add a modal filter"
-msgstr ""
+msgstr "Ajouter un filtre modal"
 
 #: src/edit/NeighbourhoodMode.svelte
 msgid "Toggle one-way (hotkey 2)"
-msgstr ""
+msgstr "Définir un sens-unique (raccourci clavier 2)"
+# TODO: here I slightly changed the meaning (define a one way)
```
dabreegster, line 1873: This seems fine. "Toggle one-way" is kind of weird in english anyway -- it's more like "toggle between one-way forwards, one-way backwards, and two-ways"

web/src/locales/fr.po
```diff
@@ -1846,201 +1852,217 @@ msgstr ""
...
 #: src/edit/NeighbourhoodMode.svelte
 msgid "Reverse directions"
-msgstr ""
+msgstr "Inverser les sens"
 
 #: src/edit/NeighbourhoodMode.svelte
 msgid "Restrict turns (hotkey 3)"
-msgstr ""
+msgstr "Restreindre les girations (raccourci clavier 3)"
+# TODO: I also noted that modal filters and one way are singular but turns and directions are plural, expected?
```
dabreegster, line 1883: ... Not really intentional :sweat_smile: Do you think it sounds better singular or plural? We could change the english ones too

dabreegster (approved):
These look great, thanks so much for starting this! I didn't notice offhand any places where the UI looks really different with shorter/longer text, but let me know if something does break.

Thanks also for the other issues filed, I'll aim to get to those later this week

PR #417, opened 2025-09-18
Started the translations, want to test them.

Kandico:


Kandico:
> These are showing up fine for me when I run locally. Awesome! Want me to merge this PR, and you continue in more followups?
>  Yes please. I will occassionally send a pull request, as running it locally from my laptop is a bit tedious and very slow. 

> How are other preparations for the use of the tool going? I'd be happy to have a quick call about it too if that'd be any easier.

We have been focusing on other projects in the last weeks, so not much progress, a quick virtual meeting next week would be really helpful next week. 

So far we have 3 goals with the ltn tool
1. Translation
2. Integrating Budapest
3. Being able to run the tool from a local server, with certain projects pre-loaded so residents can review the concepts.

dabreegster:
I think I've seen the wuchale tool automatically insert the newlines when you write a long line too. I would just write in the most convenient way for you, make sure it displays correctly, and we'll go from there

Issue #426, opened 2025-10-01
Ignore roads outside of working area


XioNoX:
I think this is a regressions as I don't think it happened before.

For example if I define my neighborhood with those boundaries : 

<img width="1353" height="1229" alt="Image" src="https://github.com/user-attachments/assets/ad8be55d-96c7-4b58-9eaa-ce34c4fa835a" />

LTN will show lots of "outbound facing" cells on the periphery : 

<img width="1353" height="1229" alt="Image" src="https://github.com/user-attachments/assets/c730ab3a-4b27-48ff-90fd-e8f1b3b39c20" />

I guess the black dashed line has something to do with it ?

XioNoX:
Thanks!

[Brest Saint-Martin-2025-10-01.geo.json](https://github.com/user-attachments/files/22636138/Brest.Saint-Martin-2025-10-01.geo.json)

To be renamed .geojson. I marked those stray roads as "primary" to workaround the issue.

For comparaison another neighborhood I'm working on doesn't have this issue (see the roads at the bottom of the working area for example): 

[Brest - Kerigonan-2025-10-01.geo.json](https://github.com/user-attachments/files/22636168/Brest.-.Kerigonan-2025-10-01.geo.json)

<img width="1460" height="1195" alt="Image" src="https://github.com/user-attachments/assets/7573faa9-f581-462e-9498-5c7aa28bc787" />

dabreegster:
Thanks for the reference files! This issue is a tricky one. There's a clear bug:

1) Draw a boundary using the snapping tool
2) It gets buffered a fair bit, picking up some stray roads sometimes
3) If you adjust boundary, you'll see the original unbuffered thing you drew. If you confirm it again (through the adjust flow), the result is only buffered by the minimal 1m and excludes stray roads.

There's a handful of places we do the buffering, depending on using the auto-generated boundaries or snapping or when merging boundaries. I will work through them carefully and figure out what's happening.

PR #428, opened 2025-10-03
Hide unimportant cells only with service roads

### LINKED ISSUE #425: Ignore dead ends cells
Not sure how best to do it, but for example if we take that private parking/driveway : https://www.openstreetmap.org/way/1412906774
It renders as its own cell, which clutters the UI.
There are at least 4 of them on that screenshot : 

Image

dabreegster:
Closes #425. @XioNoX, what do you think?

- The reasonable definition to me of "unimportant cells" is those only containing service roads. That seems to handle the cases you reported and some that I've seen elsewhere. It could get fancier by thinking about a total length of roads contained inside, or whether the cell touches multiple main roads and even has any shortcuts possible or is a true cul-de-sac. But the problem seems to come from too many small cells next to each other cluttering things, and that seems to happen around service roads.
- I've decided to make this a toggleable setting, defaulting to off (the current, cluttered behavior). It's because this does hide information that may be relevant; it is a simplification.

Before:
image
After:
image
The green and purple cells remain even after hiding. That is #426, a separate problem.

Before:
image
After:
image

I'm pretty confident about this change, so I'll merge momentarily, but feedback is very welcome and I'm happy to change this


TFCx:
> but feedback is very welcome and I'm happy to change this

I was very skeptical about this MR, which seemed strange to me. Intuitively, I would reason by coloring the roads, not the dual-surface.

But after 5/10min of brainstorming, you convinced me that if a block has only one entry/exit (unrelated to its length), it cannot serve as rat run, and thus, should not be colored distinctively. Should it the same color as its neighboring district? What if there are two neighboring districts each having its own color?

Anyway, I can see that some cluttering originates from #426.

XioNoX:
Thanks! screenshot lgtm but the change doesn't seem to be live on https://a-b-street.github.io/ltn/ ?

dabreegster:
Wait, really? I checked just now and see the new option, "Hide small cells", on the left:
image
It's not enabled by default

